PR #28, opened 2019-09-24
Return exit code from pspm_test for jenkins

eozd:
- Return exit code from pspm_test
- Run pspm_test on every push to any branch and on any pull request


eozd:
Fixes part of #29

Issue #31, opened 2019-09-25
pspm_get_events should allow choosing rising, middle or falling edge of a marker


eozd:
### Feature Description
Currently, `pspm_get_events` returns the middle of a marker impulse by default. However, this behaviour is undesirable in many situations where tagging the marker by the rising or the falling edge make more sense. Therefore, we should add an option to `pspm_get_events` so that the user can choose whether they want
1. rising edge
2. middle
3. falling edge

of a marker impulse.

In addition, this option should be cascaded all the way to the top user-level functions that at some point use `pspm_get_events`. Therefore, the main tasks are

* [x] Modify `pspm_get_events` to allow choosing the marker tag
* [x] Find top-level user space functions that depend on this functionality
* [x] Include the aforementioned option in these functions' interfaces.

irojkov-ph:
I am going through all the issues right now in preparation for the future release.

@sammaxwellxyz, if you have some time can you have a look at this please ? I think it's pretty much done except that we should be able to select the 'flank' in the GUI. Currently it's not possible. It shouldn't be too difficult I guess since `pspm_get_events` is used only by `pspm_get_markers` and `pspm_get_hb` (so only two channel type are concerned by this option). It could be nice to include it in the new release.

irojkov-ph:
This feature was implemented with #149  and added to the new release. I close it.

PR #41, opened 2019-10-09
Issue #15: pspm_get_events.m bug with 1st row markers

### LINKED ISSUE #15: pspm_get_events misses markers that correspond to the first sample row
### Summary
`pspm_get_events` is not able to locate a marker if it occurs on the first sample in a given datafile. This frequently happens with viewpoint data where the marker on the first line indicates that data recording has begun.

### Steps to Reproduce
1. Let a viewpoint datapath be stored in `fp`. Make sure that `fp` contains only one marker on the first sample line and no others. This happens quite frequently with viewpoint data.
2.  Run
```
import{1}.type = 'marker';
options.overwrite = true;
pspm_import(fp, 'viewpoint', import, options);
```

### Expected Results
No error is produced and the data marker channels is imported without any errors/warnings.

### Actual Results
A warning from `pspm_get_events.m` line 75 is issued. `lo2hi` and `hi2lo` are empty whereas they both should contain `1`, the index of the marker.

### Technical Info
* PsPM revision ID: 2a6dda7
* MATLAB version: R2018b

irojkov-ph:
Fixes #15 :

- The change is only a line (n°48) which is in a sense just avoiding the issue.
- It is detecting the first row's marker only for the `strcmpi(import.flank, 'ascending')` case otherwise it is not :(


irojkov-ph:
I am still looking for a better method than that but for this I have to speak to you for about the issue itself. Because maybe I am not understanding it correctly.

eozd:
Ok. I will check the implementation a bit myself, and we can discuss the issue on monday.

eozd:
A few comments:

1. I think using `0` in the beginning and in the end is a good idea to catch markers in the first or in the last rows. However, the definitions of other variables that are affected by this change should also be fixed. (For example `finite`)
2. As we have discussed, `viewpoint` import currently returns these `+,=,+` lines, as well. This is in my opinion a bug; therefore, I will create a new issue about it and reference it here. We will need to fix that first, and only then merge this.
3. I checked other import types, but it seems like this is more of a viewpoint related issue. Still, ensuring that a possible marker in the first or in the last line is not missed is good practice.

irojkov-ph:
Okay got it I will concentrate on #45 and look at that issue afterwards

irojkov-ph:
Shall i merge this branch now ? We will anyway modify 'pspm_get_events` according to #31.

src/pspm_get_events.m
```diff
@@ -45,7 +45,7 @@
     end
     
     % add more data in order to prevent deleting values with diff
-    data = [NaN; NaN; NaN; data; NaN; NaN; NaN;];
+    data = [0; 0; 0; data; 0; 0; 0;];
     % store information about finite and infinite in vector
     % used to reduce temp vector to relevant data
     finite = ~isnan(data);
```
eozd: Could you redefine `finite` according to the new definition of `data`?

irojkov-ph:
This is one solution that I propose, it's not ideal but at least it will catch the first and last evt.

irojkov-ph:
Yeah sorry:
- For data containing evt between `[2:end]` (i.e. no evt in `data(1)`) the program will replace `data` by `[NaN; NaN; NaN; data; 0;NaN;NaN]` (the 0 after data ensure you to catch the last evt as well if it is present otherwise it will not alter the execution of the code)

- For data having a data in the first position, the program will replace `data` by   `[NaN; head; head; data; 0;NaN;NaN]` with `head` being a random negative integer btw 0 and 1000 (I haven't set it to a precise number because i do not know what kind of evt one can have in data, if you ensure me that data is always >= 0 one could set head to -1). Then I've modified `finite` in order to avoid it taking head as a data point.

eozd:
Hmm, I am bit confused about this. I think `data` is always a boolean vector, no? In that case, why doesn't the previous solution you implemented (i.e. `data = [0;0;0;data;0;0;0];`) work, and we need something like this?

irojkov-ph:
I don't know if it is always a boolean. If yes then this solution could work: `data = [NaN; 0; 0; data; 0; 0; NaN;];`

The previous solution with `data = [0;0;0;data;0;0;0];` is not ideal because it will not catch the last evt (i.e. `data(end)`) when you use 'descending' parameter.

eozd:
It seems like it has to be a boolean array to me. Otherwise, the previous code would essentially pick every single datapoint as a marker since all the neighbouring samples in a real-valued signal are almost surely unequal. Of course this corresponds to `'continuous'` option; when it is `'timestamps'`, we don't do anything anyways.

irojkov-ph:
Okay then as said this solution is working well, should I merge ? 
(You can test if you want with the following lines and play a little bit with 'flank' parameter)
```
import.marker = 'continuous';
import.flank = 'ascending';
% import.flank = 'descending';
import.sr = 10^3;
import.data = zeros(10000,1);
idx_mrk__one = randi([1 4999],1,5);
idx_mrk__two = randi([5000 10000],1,5);
idx_mrk__three = 3001:1:3056;
import.data(idx_mrk__one) = 1;
import.data(idx_mrk__two) = 1;
import.data(idx_mrk__three) = 1;
import.data(1)=1;
import.data(end)=1;
[sts, rimport] = pspm_get_events(import);
```

PR #46, opened 2019-10-14
TODO add checkpoints to pspm_transfer_function.m 

irojkov-ph:
Changes asked by Dominik:
- Adding 3 more checkpoints in pspm_transfer_function.m which checks whenever c, Rs and offset are numeric or not.

### DIFF
--- a/src/pspm_transfer_function.m
+++ b/src/pspm_transfer_function.m
@@ -39,9 +39,9 @@ if isempty(settings), pspm_init; end;
 
 % check input arguments
 if nargin<1
-    errmsg='No data given.'; warning(errmsg); return;
+    warning('ID:invalid_input','No data given.'); return;
 elseif nargin<2
-    errmsg='No transfer constant given.'; warning(errmsg); return;
+    warning('ID:invalid_input','No transfer constant given.'); return;
 elseif ~isnumeric(c)
     warning('ID:invalid_input','The parameter ''c'' has to be numeric.'); return;
 elseif ~isnumeric(Rs)
--- a/src/pspm_transfer_function.m
+++ b/src/pspm_transfer_function.m
@@ -44,9 +44,9 @@ elseif nargin<2
     warning('ID:invalid_input','No transfer constant given.'); return;
 elseif ~isnumeric(c)
     warning('ID:invalid_input','The parameter ''c'' has to be numeric.'); return;
-elseif ~isnumeric(Rs)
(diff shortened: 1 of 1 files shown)


src/pspm_transfer_function.m
```diff
@@ -39,9 +39,15 @@
...
+    warning('ID:invalid_input','No data given.'); return;
 elseif nargin<2
-    errmsg='No transfer constant given'; warning(errmsg);
+    warning('ID:invalid_input','No transfer constant given.'); return;
+elseif ~isnumeric(c)
+    warning('ID:invalid_input','The parameter ''c'' has to be numeric.'); return;
+elseif ~isnumeric(Rs)
+    warning('ID:invalid_input','The parameter ''Rs'' has to be numeric.'); return;
+elseif ~isnumeric(offset)
+    warning('ID:invalid_input','The parameter ''offset'' has to be numeric.'); return;
```
eozd: Note that `Rs` and `offset` are optional input parameters. Therefore, we have to ensure that they were actually passed as input before performing numeric validation.

irojkov-ph: Yeah did not think about that

src/pspm_transfer_function.m
```diff
@@ -39,9 +39,15 @@
...
+    warning('ID:invalid_input','No data given.'); return;
 elseif nargin<2
-    errmsg='No transfer constant given'; warning(errmsg);
+    warning('ID:invalid_input','No transfer constant given.'); return;
+elseif ~isnumeric(c)
+    warning('ID:invalid_input','The parameter ''c'' has to be numeric.'); return;
+elseif exist('Rs','var') && ~isnumeric(Rs)
+    warning('ID:invalid_input','The parameter ''Rs'' has to be numeric.'); return;
+elseif exist('offset','var') && ~isnumeric(offset)
+    warning('ID:invalid_input','The parameter ''offset'' has to be numeric.'); return;
```
eozd: This works; but it makes the logic a bit hard to follow. I think it would be easier to maintain if we perform the argument validation after the `nargin` checks. This has two advantages:
1. We don't have to check for existence, because any variable that was not passed is already initialized with some default value.
2. `nargin` checks already do "number of parameters checks". Adding extra `exists` logic inside this if-elseif block makes the logic hard to follow. Avoiding this is beneficial for later debugging.

irojkov-ph: Fair enough, it makes more sense !

PR #47, opened 2019-10-16
Fixes issue #45: changes in import_viewpoint function

### LINKED ISSUE #45: pspm_get_viewpoint returns +,=,+ lines in the marker list
### Summary
`pspm_get_viewpoint` returns `+,=,+` lines, which denote the beginning of data recording, in the marker list. Since these are not important for researchers, these lines should be filtered out of the marker list.

### Steps to Reproduce
1. Import the marker channel from a viewpoint file via
```MATLAB
import{1}.type = 'marker';
[sts, out, srcinfo] = pspm_get_viewpoint(datafile_path, import);
```

### Expected Results
Marker channel contains only the markers entered by the user.

### Actual Results
Marker channel contains `+,=,+` lines, as well.

We need to filter markers that denote the beginning of data recording out of the marker list. However, this should be done intelligently so that no actual marker is filtered out.

Currently `import_viewpoint` function which handles the datafile reading part is doing its part correctly by returning all the markers in the file. However, `pspm_get_viewpoint` should perform this filtering so that the returned marker channel conforms to PsPM standard.

### Technical Info
* PsPM revision ID: 0332221
* MATLAB version: R2018b

irojkov-ph:
Fixes #45 .

- Suppress all non-markers from the markers arrays in the `import_viewpoint` function.

### DIFF
--- a/src/Import/viewpoint/import_viewpoint.m
+++ b/src/Import/viewpoint/import_viewpoint.m
@@ -49,6 +49,13 @@ function [data] = import_viewpoint(filepath)
     [channels, marker, chan_info] = parse_messages(messages, channels, marker, chan_info, file_info.eyesObserved);
 
     sess_beg_indices = [find(contains(marker, '+')); size(dataraw, 1) + 1];
+
+    % After setting the beginning index we delete all ',','+','=' form markers array
+    NotMarkers = [',','+','=']; 
+    for nmkr=1:size(NotMarkers,2)
+       marker = strrep(marker,NotMarkers(nmkr),'');
+    end
+
     for sn = 1:numel(sess_beg_indices) - 1
         begidx = sess_beg_indices(sn);
         endidx = sess_beg_indices(sn + 1) - 1;
--- a/src/Import/viewpoint/import_viewpoint.m
+++ b/src/Import/viewpoint/import_viewpoint.m
@@ -49,13 +49,7 @@ function [data] = import_viewpoint(filepath)
     [channels, marker, chan_info] = parse_messages(messages, channels, marker, chan_info, file_info.eyesObserved);
 
     sess_beg_indices = [find(contains(marker, '+')); size(dataraw, 1) + 1];
-
-    % After setting the beginning index we delete all ',','+','=' form markers array
-    NotMarkers = [',','+','=']; 
-    for nmkr=1:size(NotMarkers,2)
-       marker = strrep(marker,NotMarkers(nmkr),'');
-    end
-
+    
     for sn = 1:numel(sess_beg_indices) - 1
         begidx = sess_beg_indices(sn);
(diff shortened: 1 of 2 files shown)


src/Import/viewpoint/import_viewpoint.m
```diff
@@ -49,6 +49,13 @@
     [channels, marker, chan_info] = parse_messages(messages, channels, marker, chan_info, file_info.eyesObserved);
 
     sess_beg_indices = [find(contains(marker, '+')); size(dataraw, 1) + 1];
+
+    % After setting the beginning index we delete all ',','+','=' form markers array
+    NotMarkers = [',','+','=']; 
+    for nmkr=1:size(NotMarkers,2)
+       marker = strrep(marker,NotMarkers(nmkr),'');
+    end
+
```
eozd: 1. I believe that this filtering should be implemented in `pspm_get_viewpoint` function, not in `import_viewpoint`. There are two reasons for this:
  * It is highly likely that doing this kind of filtering in `import_viewpoint` itself will require modifications to other sections, as well. Therefore, it will require more changes than this.
  * It is against the general rationale behind `import` and `get` function separation. `import` functions are intended to return whatever is in the datafile in some sort of structured format, whereas `get` functions are intended to process this and make it compatible with PsPM functions.
2. Note that while filtering the markers according to name, you should also get rid of corresponding timesteps.
3. We need to do the filtering a bit more strictly since these characters may also appear in proper marker names such as `CS+`. For filtering the name, I would suggest removing the markers whose names consist of only `,+=` characters.

irojkov-ph:
Can you review these changes, my 'filter' captures the 'CS+' in '+,=,=,+,=,=,==,CS+'

src/pspm_get_viewpoint.m
```diff
@@ -309,6 +309,18 @@
 end
 
 function import_cell = import_marker_chan(import_cell, markers, mi_names, mi_values, n_rows, sampling_rate)
+    % Put here all characters which do not belong to markers
+    % they have to be separated by a '|'
+    % '=' has a special threatment because we do not want to delete 'CS+'
+    non_markers = [',','|','(?<!\w)(+)','|','='];
+    
+    mi_names = regexprep(mi_names,non_markers,'');
+    non_empty = find(~cellfun('isempty',mi_names));
```
eozd: This would work, I think. We don't even need the complex regex pattern for `+` since if the marker has at least one character different than `,+=`, it will still be non-empty which is enough. However, it would be better to assign the regex result to a new variable in line 317 instead of `mi_names`. The reason is that currently this can change the name of a marker. For example, if a marker is named `CS=`, this would return only `CS`.

eozd: As a side note, I noticed a bug in import regarding session beginning offsets. I implemented a quick fix and will submit it once we fix this one.

PR #49, opened 2019-10-21
Issue48 hb2hp bug

### LINKED ISSUE #48: pspm_hb2hp do not take into account entered options
### Summary
`pspm_hb2hp` do not take into account entered options such as the upper and lower limits and `channel_action`. The functions takes always the default parameters.

### Steps to Reproduce
1. % Download the file
[test_hb2hp_data2.zip](https://github.com/bachlab/PsPM/files/3750163/test_hb2hp_data2.zip)
2. option.channel_action = 'abc';
3. pspm_hb2hp('test_hb2hp_data2.mat',100,[],option);

### Expected Results
An error, because there is no such saving option in the `pspm_write_channel`, it has to be either `add` or `replace` instead of `abc` on the line 2.

### Actual Results
The function will produce `ans = 1`.

The cause is that `pspm_hb2hp` function replaces all the given options by the default ones.

### Technical Info
* PsPM revision ID: 76a3927ac1d8de58e1f21a25e696046003c220b8
* MATLAB version: R2019b

irojkov-ph:
Fixes #48 .

- `try options.channel_actions catch ... ` is an always true action, that means it will always catch, that's why I changed it to `if` statements.

### DIFF
--- a/src/pspm_hb2hp.m
+++ b/src/pspm_hb2hp.m
@@ -35,10 +35,11 @@ sts = -1;
 global settings;
 if isempty(settings), pspm_init; end;
 
-try option.channel_action; catch options.channel_action = 'replace'; end;
-try options.limit; catch options.limit = struct(); end;
-try options.limit.upper; catch options.limit.upper = 2; end;
-try options.limit.lower; catch options.limit.lower = 0.2; end;
+if ~exist('options','var'), options = struct(); end;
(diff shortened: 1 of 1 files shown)


eozd:
`try catch` is actually a valid technique, and it works. The problem in this case was that in the first line, variable `option` is checked instead of `options`. But this newer version also does the job and fixes the issue.

irojkov-ph:
Oh shit sorry I haven’t seen it 😅 thank you

PR #54, opened 2019-10-22
Check if a marker only contains one of +,= for splitting

### LINKED ISSUE #53: import_viewpoint creates a new session with CS+ marker
### Summary
`import_viewpoint` creates a new session if a marker contains `+` somewhere. Therefore a `CS+` marker causes a session split.

### Steps to Reproduce
1. Create a viewpoint datafile containing a `CS+` marker and only one session
2. Run `import_viewpoint`
3.

### Expected Results
Only one session is returned

### Actual Results
Multiple sessions are returned

Add an extra check to ensure a marker contains only `+=,` characters.

### Technical Info
* PsPM revision ID: b25ec0b
* MATLAB version: R2018b

eozd:
Fixes #53 .

- Add extra check to "new session marker" detection: a marker should only contain `+,=` characters.

### DIFF
--- a/src/Import/viewpoint/import_viewpoint.m
+++ b/src/Import/viewpoint/import_viewpoint.m
@@ -48,7 +48,17 @@ function [data] = import_viewpoint(filepath)
     channels = dataraw(:, chan_info.col_idx);
     [channels, marker, chan_info] = parse_messages(messages, channels, marker, chan_info, file_info.eyesObserved);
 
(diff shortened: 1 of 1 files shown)


eozd:
Yes, I see the point. But such a marker is not allowed for a start marker. If there is such a marker in the file that denotes a session starting, the file is syntactically incorrect imo.

irojkov-ph:
Okay now I've understood what you mean ! Then every thing is all right ;)

Issue #37, opened 2019-10-07
Implement tests for frequently used PsPM functions


eozd:
### Summary
Below functions implement some core functionality, but currently are not tested. We need to implement tests for them.

- [x] `pspm_hb2hp.m`
- [x] `pspm_filtfilt.m`
- [x] `pspm_butter.m`

irojkov-ph:
Should I also modify the Developer Guide ?

eozd:
Yes, please. We are all set for the new release, so we should prepare the manual and developer guide as the first step.

PR #55, opened 2019-10-23
TODO: test for pspm_hb2hp.m

irojkov-ph:
Fixes partially #37  .

- In `pspm_hb2hp.m`, I've modified a little bit warning messages.
- Created a test file for testing that function.
- It needs two files which I've attached here and which are supposed to go in the `ecg2hb`folder.
[test_data.zip](https://github.com/bachlab/PsPM/files/3762513/test_data.zip)

### DIFF
--- a/src/pspm_hb2hp.m
+++ b/src/pspm_hb2hp.m
@@ -94,7 +94,8 @@ newdata.header.chantype = 'hp';
 
 o.msg.prefix = 'Heart beat converted to heart period and';
 try
-    [~,winfos] = pspm_write_channel(fn, newdata, options.channel_action, o);
+    [nsts,winfos] = pspm_write_channel(fn, newdata, options.channel_action, o);
+    if nsts == -1, return; end
 catch
     warning('ID:invalid_input', 'call of pspm_write_channel failed');
     return; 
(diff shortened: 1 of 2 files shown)


src/pspm_hb2hp.m
```diff
@@ -93,8 +93,9 @@
 newdata.header.chantype = 'hp';
 
 o.msg.prefix = 'Heart beat converted to heart period and';
-[sts, winfos] = pspm_write_channel(fn, newdata, options.channel_action, o);
-if nsts == -1
+try
+    [~,winfos] = pspm_write_channel(fn, newdata, options.channel_action, o);
+catch
```
eozd: I assume previously it was a typo to check `nsts == -1`. I am in favor of using try catch here, but we still should check for the status flag of `pspm_write_channel`. Maybe put an assertion after calling `pspm_write_channel` inside the try block?

eozd: Apart from this, I think it looks good. Feel free to merge it afterwards.

irojkov-ph: Got it, thanks for the review !

PR #57, opened 2019-10-28
Created a warning

### LINKED ISSUE #56: pspm_filtfilt: Invalid filtering parameters
### Summary
When we execute `pspm_filtfilt` with filtering parameters a and b of dimension 1 (i.e. numbers) one get an error of exceeding index.

### Steps to Reproduce
1. pspm_filtfilt(1,1,[1:10]);

### Expected Results
A warning should appear.

### Actual Results
The following error occurs:
`Index exceeds the number of array elements (1).`

The cause is the formula we use which is given on lines 12 to 15 of the `pspm_filtfilt.m`:
```
% The filter is described by the difference equation:
%
%     y(n) = b(1)*x(n) + b(2)*x(n-1) + ... + b(nb+1)*x(n-nb)
%                      - a(2)*y(n-1) - ... - a(na+1)*y(n-na)
```
The solution would be to either modify the formula or to throw a warning. Because of the fact that this function is used a lot the easiest way is to make an assertion and throw a warning.

### Technical Info
* PsPM revision ID: b5ec33af6fa995f3704051c30d1466ba72b3c5d6
* MATLAB version: R2019b

irojkov-ph:
Fixes #56  .

- Created a warning

### DIFF
--- a/src/pspm_filtfilt.m
+++ b/src/pspm_filtfilt.m
@@ -70,8 +70,8 @@ nfact = 3*(nfilt-1);
 if len <= nfact
     error('Data must have length more than 3 times filter order.');
 end
-if na == 1
-    warning('ID:invalid_input','The filter parameters ''a'' must be of dimension 2.'); return;
+if nfilt == 1
+    y=x; return
 end
 
 % Use sparse matrix to solve system of linear equations for initial 
--- a/src/pspm_filtfilt.m
+++ b/src/pspm_filtfilt.m
@@ -70,6 +70,9 @@ nfact = 3*(nfilt-1);
 if len <= nfact
     error('Data must have length more than 3 times filter order.');
 end
+if na == 1
+    warning('ID:invalid_input','The filter parameters ''a'' must be of dimension 2.'); return;
+end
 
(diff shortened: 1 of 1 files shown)


src/pspm_filtfilt.m
```diff
@@ -70,6 +70,9 @@
 if len <= nfact
     error('Data must have length more than 3 times filter order.');
 end
+if na == 1
+    warning('ID:invalid_input','The filter parameters ''a'' must be of dimension 2.'); return;
+end
```
eozd: At this point it is possible that length of `a` has increased due to line 66. Therefore even if `na == 1`, it is possible that the code below would work. I think it would be better if we rather verify that `nfilt` is at least 2, which corresponds to having a filter of order at least 1. If it doesn't hold then we can emit a warning, saying that the parameter vectors are too short.

eozd: So after the discussion in #58, we want to check if `nfilt == 1`, and return the signal unchanged if it is. All the other cases seem appropriately handled to me. What do you think?

irojkov-ph: Yes for me too. I changed and merged that.

PR #58, opened 2019-10-28
TODO: test for pspm_filtfilt.m

irojkov-ph:
Fixes partially #37 .

- Modified a little bit warning messages.
- Created a test file for testing that function (it needs the pull request #57 to work properly)


fmelinscak:
> Apparently we need to handle the case when the lengths of both `a` and `b` are 1 in a special way; so let's not merge this one just yet.

Yes, I would suggest to keep the same behavior as the Matlab function `filtfilt`, after which `pspm_filtfilt` is modeled. In this case, the signal is actually returned unchanged. However, more generally the case where `length(a)` is 1, should be handled appropriately.

PR #65, opened 2019-11-04
Extend get_events bugfix according to the previously implemented tests

eozd:
- In #15 we fixed the issue regarding missing the first marker in a file. However, our implementation misses a few cases that were tested in `pspm_get_events_test`. This pull request handles these issues. The main idea of the proposed changes are as below:
  * We missed two usecases: the input signal may be shifted by a baseline and the pulses might also be triangular. To handle both of these cases with the new code, there is a new signal normalization portion added.
  * Since we use a normalized signal for detection, the original signal is stored in an extra variable in order to return correct marker values later in the code.

### DIFF
--- a/src/pspm_get_events.m
+++ b/src/pspm_get_events.m
@@ -44,6 +44,13 @@ elseif strcmpi(import.marker, 'continuous')
         data = data';
     end
     
+    possible_values = unique(data);
+    min_values_indices = data == min(possible_values);
+    max_values_indices = data == max(possible_values);
+    data_orig = data;
+    data = (data + min(possible_values)) / (max(possible_values) - min(possible_values));
+    data(min_values_indices) = 0;
+    data(max_values_indices) = 1;
     % add more data in order to prevent deleting values with diff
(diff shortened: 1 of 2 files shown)


src/pspm_get_events.m
```diff
@@ -108,7 +115,7 @@
         baseline = v(end, 1);
         
         % we are interested in the delta -> remove "baseline offset"
-        values = data(round(mPos)) - baseline;
+        values = data_orig(round(mPos) - 3) - baseline;
```
eozd: I am not really happy by this `-3` part. Maybe there is a better way to implement it?

irojkov-ph: What this 3 is doing here ? Why do we need it ?

eozd:
> What this 3 is doing here ? Why do we need it ?

So `mPos` is defined using `data` which has 3 extra zeroes padded at the front. Therefore, its indices will be offset by 3

irojkov-ph:
Ok I see, fair enough. For now I think it is enough and for 4.3.0 we will anyway modify this function.

PR #73, opened 2019-11-20
Fix/issue67 : New procedure for catching markers in an Eyelink file

### LINKED ISSUE #67: Number of marker names and marker onsets mismatch in Eyelink import
### Steps to Reproduce
1. Import an Eyelink `.asc` file containing some markers using `pspm_get_eyelink` function
2. Eyelink data file should contain markers with the following properties:
  * there is at least one marker whose timestamp is outside the timestamp range of the data samples
  * there are at least two markers whose timestamps are exactly the same

### Actual Results
```MATLAB
struct with fields:
data: [2858×1 double]
markerinfo: [1×1 struct]
header: [1×1 struct]

        data{2}.markerinfo
        ans =
        struct with fields:
        name: {2860×1 cell}
        value: [2860×1 double]
```

I couldn't find anything in the pspm_trim.m file that takes into account these size differences. My guess is that it has something to do with the import_eyelink.m function and that some irrelevant eyelink messages are saved as marker names. However, I have not had the time to verify this with a different data set.

PSPM v4.2.0
import_eyelink.m
line 259
```MATLAB
% construct markers
markers.markers = false(size(dataraw, 1), 1);
markers.times = [];
markers.names = {};
for idx = msg_indices
    msgline = messages{idx};
    parts = split(msgline);
    time = str2num(parts{2});
    markers.markers(bsearch(timecol, time)) = true;
    markers.times(end + 1, 1) = time;
    markers.names{end + 1, 1} = cell2mat(join(parts(3:end), ' '));
end
```

> My work around

```MATLAB
% construct markers
markers.markers = false(size(dataraw, 1), 1);
markers.times = [];
markers.names = {};
for idx = msg_indices
    msgline = messages{idx};
    parts = split(msgline);
    time = str2num(parts{2});
    % I have several cases where 2 markers times are identical although with
    % different names. Identical times become problematic in the next
    % step when we use them as index. This is my workaround so far...
    if ~isempty(markers.times) && time == markers.times(end)
        time = time+1;
    end
    % I am using ismeber instead of index = bsearch(x,var) here because
    % if the var that we are looking for is not included in the data vector
    % then index becomes max(x).
    i_search = find(ismember(timecol, time));
    if ~isempty(i_search)
        markers.markers(i_search) = true;
        markers.times(end + 1, 1) = time;
        markers.names{end + 1, 1} = cell2mat(join(parts(3:end), ' '));
    else
        %dont save time and name
    end
end
```
"""

### Technical Info
* Reference issue (original): https://sourceforge.net/p/pspm/tickets/17 
* PsPM revision ID: PsPM 4.2.0
* MATLAB version: R2017b

irojkov-ph:
Fixes #67 .

Changes proposed in this pull request (per file):
1. `import_eyelink.m` : 
     - implement a condition that does not allow us to import markers which are outside the session end time (i.e. the last timestamp before the marker `END`)
2. `pspm_get_events.m` :
     - added a new `import.flank` parameter called `'all'` which imports all markers and data related to them.
     - `pspm_get_eyelink` function calls `pspm_get_events` in order to transform the original continuous markers data to an vector of "events", but when it encounter data such as the following one it will consider consecutive markers as one continuous one and thus with the option `ascending` it will catch the timestamp only of the first one.
```
------ Raw Eyelink data -------    ------ Data that pspm_get_events receives -------
6084232	   .	   .	    0.0            6084232        0
6084233	   .	   .	    0.0            6084233        0
6084234	   .	   .	    0.0            6084234        0
MSG	6084235 Start End                  6084235        1
6084235	   .	   .	    0.0                                                           
MSG	6084236 JitterFix                  6084236        1
6084236	   .	   .	    0.0                                                        
6084237	   .	   .	    0.0            6084237        0
6084238	   .	   .	    0.0            6084238        0
6084239	   .	   .	    0.0            6084239        0
```
3. `pspm_get_eyelink.m` :
     - I have changed the `import.flank` parameter to `'all'` for eyelink files according to the changes from point 2.

### DIFF
--- a/src/Import/eyelink/import_eyelink.m
+++ b/src/Import/eyelink/import_eyelink.m
@@ -299,9 +299,11 @@ function [dataraw, markers, chan_info] = parse_messages(messages, dataraw, chan_
         msgline = messages{idx};
         parts = split(msgline);
         time = str2num(parts{2});
-        markers.markers(bsearch(timecol, time)) = true;
-        markers.times(end + 1, 1) = time;
-        markers.names{end + 1, 1} = cell2mat(join(parts(3:end), ' '));
+        if time < session_end_time
+            markers.markers(bsearch(timecol, time)) = true;
+            markers.times(end + 1, 1) = time;
+            markers.names{end + 1, 1} = cell2mat(join(parts(3:end), ' '));
+        end
     end
 
     % set data columns
--- a/src/Import/eyelink/import_eyelink.m
+++ b/src/Import/eyelink/import_eyelink.m
@@ -299,7 +299,7 @@ function [dataraw, markers, chan_info] = parse_messages(messages, dataraw, chan_
         msgline = messages{idx};
         parts = split(msgline);
         time = str2num(parts{2});
-        if time < session_end_time
+        if time <= session_end_time
             markers.markers(bsearch(timecol, time)) = true;
             markers.times(end + 1, 1) = time;
             markers.names{end + 1, 1} = cell2mat(join(parts(3:end), ' '));
--- a/src/pspm_get_events.m
+++ b/src/pspm_get_events.m
@@ -6,7 +6,7 @@ function [sts, import] = pspm_get_events(import)
 %                  .marker ('timestamps', 'continuous')
 %                  .sr (timestamps: timeunits in seconds, continuous: sample rate in 1/seconds)
 %                  and optional field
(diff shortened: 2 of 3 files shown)


eozd:
`import_eyelink` test seems to be failing, even though Jenkins returned success for some reason. I will check the jenkins part; could you in the meantime fix the issue with `import_eyelink` test @irojkov-ph  ?

Issue #50, opened 2019-10-21
Estimating inputs to the pupil model


fmelinscak:
### Feature Description

I'm relaying a feature request from @dominikbach: the feature in question is estimating inputs of the pupil response model.
The idea is to enable analyses similar to ones from the [Korn & Bach (2016) paper](https://www.doi.org/10.1167/16.3.28) (in particular, analyses leading to figures 9, 10, 11).

The assumption here is that the pupil can be modeled as an LTI system, with the exact response function depending on whether pupil dilations or constrictions are modeled (this is implemented in PsPM functions `pspm_bf_ldrf_gm.m` and `pspm_bf_lcrf_gm.m` respectively).
Additionally, the method assumes the functional form of the input signal (in the paper, this is a gamma pdf), and the parameters of this input function are optimized such that the squared error between the prediction (convolution of the input and the LTI response function) and the observed data is minimized.
The paper doesn't explicitly mention which optimization method was used, but I think it might have been `fminsearch` (other analyses in the paper use this method).

The function should have a similar API as `pspm_glm`, and it should support condition-wise model fitting.
Importantly (and differently from the GLM), all the responses of a condition should be averaged before the input is estimated.

To start work on this, you would need some pupil size data with trial-wise markers.
I believe Dominik will get in touch with Christoph about getting the data from the paper, but in the meantime, you can use any pupil size data that has markers.

eozd:
It seems that this new function shares many of the required functionalities with `pspm_glm`. Do we want some sort of refactoring of `pspm_glm` before starting the implementation?

fmelinscak:
As a user, I don't have a strong opinion either way. But I think it would be ok to refactor `pspm_glm` only if there are comprehensive tests in place (and, in particular, regression tests with some actual data), to ensure that results of `pspm_glm` don't change due to refactoring, since it's a workhorse of PsPM-based analysis pipelines.

PR #83, opened 2020-01-30
Improvement for pspm_extract_segments

irojkov-ph:
Here I want to propose a Improvement for `pspm_extract_segments` function which will be very useful for the feature request #50 and could be also for other future functions.

- The main change is that now with these changes one can use the functions as following:
``` [sts, segments] = pspm_extract_segments('manual', data_raw, sr, timing, options)```
where `data_raw` would be a numeric array of a cell array of numeric arrays of raw data and `sr`  the sampling rate of the data (or an array of sampling rates of data).
- The test for this function will not work because of the line n° 184 which is now a valid input for the function. Thus it has to be replaced by the following lines:
```
this.verifyWarning(@() pspm_extract_segments('manual',struct('a',10),0,timing), 'ID:invalid_input');
this.verifyWarning(@() pspm_extract_segments('manual',[1,3],logical(32),timing), 'ID:invalid_input');
```
All the rest of the tests stays valid.
- I have tested the new function with my data and it reproduces the result in an identical way as when it was used with the previous version of `pspm_extract_segments`.

### CODE REVIEW

**Comment by irojkov-ph** [comment]
In the last commit I've change the function such that "timeunit" should be seconds when raw data are given to it. Obviously markers and other time unit wouldn't make sense since we do not have any filename.

**Comment by irojkov-ph** [comment]
An update of this pull request:
- I've made some changes in the original code (typically on lines 569-587) because when making deeper test of the actual function in the manual mode w/o raw data, I've realized that there are some errors. Thus I've corrected it.
- Secondly, upon reflection and after discussing it with @fmelinscak, I've changed my initial implementation. Now one can choose any of the three types of time units when using this function with raw data.

File: src/pspm_extract_segments.m [hunk 1]
```diff
@@ -1,680 +1,738 @@
...
+    %                           will be treated as session. Onset values are
+    %                           averaged through conditions and sessions.
+    %       data_raw:           Numeric raw data or a cell array of numeric raw data.
+    %       chan:               Channel number or cell of channel numbers which
+    %                           defines which channel should be taken to
+    %                           extract the segments. Chan should correspond to
+    %                           data_fn and should have the same length. If
+    %                           data_fn is a cell and chan is a single number,
+    %                           the number will be taken for all files.
+    %       sr:                 Array of sampling rates (same dimension as data_raw)
```
eozd: Writing "same dimension as `data_raw`" might be confusing, so we should note that this is the case only if `data_raw` is a cell array.

src/pspm_extract_segments.m
```diff
@@ -1,680 +1,738 @@
...
+    if ~isfield(options, 'outputfile')
+        options.outputfile = '';
+    end;
+    
+    % overwrite
+    if ~isfield(options, 'overwrite')
+        options.overwrite = 0;
+    end;
+    
+    % dont_ask_overwrite
```
eozd: (Low priority) Now I see that we have this undocumented previously implemented "don't ask overwrite" option for some reason, and we don't need it. Could you remove everything related to this option and change line 715 to `else`?

eozd:
Also, could you fix the failing test? If this is related to the comment you made about tests, please do change the tests in this pull request as well so that everything passes; afterwards I will approve it.

src/pspm_extract_segments.m
irojkov-ph: Copy that

src/pspm_extract_segments.m
irojkov-ph: Copy that

PR #87, opened 2020-02-12
Feature82/pspm display

### LINKED ISSUE #82: Increase available y-axis units in pspm_display
### Feature Description
The y-axis name and units are set in `pspm_display` on lines 773-786 with the following block of code. For now it treats only few cases. Thus, this list should be increased to the most used data types. (Dominik's request)

```
wv_chanid = handles.prop.wavechans(handles.prop.idwave);
unit = deblank(handles.data{wv_chanid}.header.units);
if strcmp(handles.prop.wave,'ecg')
   ylabel([' Amplitude [',unit,'] '],'Fontsize',14);
elseif strcmp(handles.prop.wave,'scr')
   ylabel([' Amplitude [',unit,'] '],'Fontsize',14)
elseif strcmp(handles.prop.wave,'emg')
   ylabel([' Amplitude [',unit,'] '],'Fontsize',14)
elseif strcmp(handles.prop.wave,'hp')
   ylabel(' Interpolated IBI [ms] ','Fontsize',14)
elseif strcmp(handles.prop.wave,'pupil')
   ylabel(' size in arbitrary units ','Fontsize',14)
else ylabel([' unknown unit [',unit,'] '],'Fontsize',14)
end 
```

irojkov-ph:
Fixes #82 .

- I've modified the pspm_display file to take into account the pupil size units and the gaze x & y coordinates. I've made a distinction between x and y coordinates but not between left and right. If I should do it just tell me that in the comments and I'll modify it ;)
- I've kept lines 783-784 (see below) but not sure if it is still necessary since (as I know) it was the previous format and that now we record data directly in correct units and in channels "pupil_l" and "pupil_r".
```
elseif strcmp(handles.prop.wave,'pupil')
            ylabel(' size in arbitrary units ','Fontsize',14)
```

### DIFF
--- a/src/pspm_display.m
+++ b/src/pspm_display.m
@@ -782,6 +782,12 @@ elseif not(isempty(marker)) || not(isempty(wave)) || not(isempty(hbeat))
             ylabel(' Interpolated IBI [ms] ','Fontsize',14)
         elseif strcmp(handles.prop.wave,'pupil')
             ylabel(' size in arbitrary units ','Fontsize',14)
+        elseif contains(handles.prop.wave,'pupil')
+            ylabel([' Pupil size [',unit,'] '],'Fontsize',14)
+        elseif contains(handles.prop.wave,'gaze') && contains(handles.prop.wave,'x')
+            ylabel([' Gaze x coordinate [',unit,'] '],'Fontsize',14)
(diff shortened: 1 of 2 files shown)


irojkov-ph:
Merge after #83

irojkov-ph:
It says that there are chages in pspm_extract_segments but I revert them like that this pull request concerns only the pspm_display

PR #92, opened 2020-02-17
Feature78/sps spl

### LINKED ISSUE #78: Create a pspm_find_scanpaths
### Feature Description
Feature requested by Yanfang. The goal is to add a new function to PsPM which would extract the scanpath length from pupil data based on the work she has done in her paper. It includes to:
 - [ ] Check if `pspm_bf_spsrf_box.m` and `pspm_bf_spsrf_gamma.m` work properly (they are needed in the function)
 - [ ] Create the new function 
 - [ ] In `pspm_init.m` one has to add a new channel type + add a new `default.glm` parameter.

irojkov-ph:
Fixes #78 .

As commented in the issue, the feature was already implemented in PsPM by Laure but after checking this implementation using Yanfang's dataset I've realized that there are some errors and inconsistencies in the code
- `pspm_compute_visual_angle` has nothing to do with `pupil` data type and presence of an error in the longitude/latitude calculation.
- `pspm_convert_pixel2unit` had an error in the conversion factor of pixels wrt to mm (can you please confirm that) and also an inconsistency in the way it called the `pspm_compute_visual_angle`.

### DIFF
--- a/src/pspm_compute_visual_angle.m
+++ b/src/pspm_compute_visual_angle.m
@@ -105,7 +105,11 @@ for i=1:n_eyes
             % get channel specific data
             gx_d = data{gx}.data;
             gy_d = data{gy}.data;
-%             gy_d = data{gy}.header.range(2)-gy_d;                         <----- need an explanation for that
+            
+            % The convention is that the origin of the screen is in the bottom 
+            % left corner, so the following line is not needed a priori, but I 
+            % leave it anyway just in case :
+            % gy_d = data{gy}.header.range(2)-gy_d; 
 
             N = numel(gx_d);
             if N~=numel(gy_d)
--- a/src/pspm_compute_visual_angle.m
+++ b/src/pspm_compute_visual_angle.m
@@ -1,7 +1,9 @@
 function [sts, out] = pspm_compute_visual_angle(fn,chan,width,height, distance,unit,options)
 % pspm_compute_visual_angle computes from gaze data the corresponding
-% visual angle (for each data point)
-
+% visual angle (for each data point). The convention used here is that the
+% origin of coordinate system for gaze data is at the bottom left corner of
+% the screen.
+% 
 % FORMAT:
 %        [sts, out] = pspm_compute_visual_angle(fn,chan,width,height, distance,unit,options)
 %
@@ -122,19 +124,19 @@ for i=1:n_eyes
             gy_d = gy_d - height/2;
 
             %compute visual angle for gaze_x and gaze_y
-            s_x = gx_d; % x axis in spherical coordinates
-            s_y = distance * ones(numel(gx_d),1);% y axis in spherical coordinates, actually is the radial from participant to the screen
-            s_z = gy_d;  % z axis in spherical coordinates, actually is y axis on the screen
-            [azimuth, elevation, r]= cart2sph(s_x,s_y,s_z);% convert cartesian to spherical coordinates in radians, azimuth = longitude, elevation = latitude
+            s_x = gx_d;                                     % x axis in cartesian coordinates
+            s_y = distance * ones(numel(gx_d),1);           % y axis in cartesian coordinates, actually the distance from participant to the screen
+            s_z = gy_d;                                     % z axis in spherical coordinates, actually the y axis of the screen
+            [azimuth, elevation, ~]= cart2sph(s_x,s_y,s_z); % convert cartesian to spherical coordinates in radians, azimuth = longitude, elevation = latitude
 
-            lat = rad2deg(elevation);% convert radians into degrees
+            lat = rad2deg(elevation);   % convert radians into degrees
             lon = rad2deg(azimuth);
 
             %compute visual angle for range 
             r_x = [-width/2,width/2,0,0]';
             r_y = distance * ones(numel(r_x),1);
             r_z = [0,0,-height/2,height/2]';
-            [x_range_sp, y_range_sp,r_range]= cart2sph(r_x,r_y,r_z);
+            [x_range_sp, y_range_sp,~]= cart2sph(r_x,r_y,r_z);
             x_range_sp = rad2deg(x_range_sp);
             y_range_sp = rad2deg(y_range_sp);
 
--- a/src/pspm_compute_visual_angle.m
+++ b/src/pspm_compute_visual_angle.m
@@ -20,13 +20,13 @@ function [sts, out] = pspm_compute_visual_angle(fn,chan,width,height, distance,u
 %       distance:           distance between eye and screen in length units.
 %       unit:               unit in which width, height and distance are given.
 %       options:            
-%         - channel_action: ['add'/'replace'] Defines whether the new channels
+%         .channel_action:  ['add'/'replace'] Defines whether the new channels
 %                           should be added or the previous outputs of this function
 %                           should be replaced.
 %                           Default: 'add'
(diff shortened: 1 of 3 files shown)


src/pspm_convert_pixel2unit.m
```diff
@@ -161,7 +161,13 @@
 
 if strcmpi(unit,'degree')
     options.channel_action = 'replace';
-    [lsts, outinfo] = pspm_compute_visual_angle(fn,0,width, height, distance,unit_h_w_d,options);
+    [~, lsts, outinfo] = evalc(['pspm_compute_visual_angle(fn,outinfo.channel,'...
```
eozd: I agree with passing `outinfo.channel` as a parameter, but I don't think we need to use `evalc`. Why not just call the function directly?

src/pspm_convert_pixel2unit.m
```diff
@@ -135,11 +135,11 @@
         
         % convert according to range
         chan.data = (chan.data-chan.header.range(1)) ...
-            ./ diff(chan.header.range) * fact;
+                    * fact ./ (diff(chan.header.range) + 1);
```
eozd: Not sure why we have `+ 1` now? Because of a possible division by 0? If so, that should be handled by a separate if check instead of changing the result of the operation.

src/pspm_compute_visual_angle.m
```diff
@@ -93,92 +89,75 @@
     if strcmpi(options.eyes, 'all') || strcmpi(options.eyes(1), eye)
```
eozd: If I am not mistaken, `infos.source.eyesObserved` would contain one of `l`, `r` or `lr`. So the `strcmpi` check against `left`, `right` or `all` won't work. Could you fix this naming discrepancy?

src/pspm_compute_visual_angle.m
```diff
@@ -93,92 +89,75 @@
     if strcmpi(options.eyes, 'all') || strcmpi(options.eyes(1), eye)
         gaze_x = ['gaze_x_', eye];
```
eozd: So `eye` variable here should be `l` or `r` since the name of the channel can only be `gaze_x_l`, gaze_y_r`, etc.

src/pspm_compute_visual_angle.m
```diff
@@ -93,92 +89,75 @@
...
-                lon = rad2deg(azimuth);
-                
-                %compute visual angle for range 
-                r_x = [-width/2,width/2,0,0]';
-                r_y = distance * ones(numel(r_x),1);
-                r_z = [0,0,-height/2,height/2]';
-                [x_range_sp, y_range_sp,r_range]= cart2sph(r_x,r_y,r_z);
-                x_range_sp = rad2deg(x_range_sp);
-                y_range_sp = rad2deg(y_range_sp);
+            % The convention is that the origin of the screen is in the bottom
```
eozd: I think we should mention this assumption in the function documentation, because other eye trackers may be using different conventions.

src/pspm_compute_visual_angle.m
```diff
@@ -93,92 +89,75 @@
...
-                % longitude (azimuth angle from positive x axis in horizontal plane) of gaze points
-                visual_angl_chans{p}.data = lon;
-                visual_angl_chans{p}.header.units = 'degree';
-                visual_angl_chans{p}.header.range = [x_range_sp(1),x_range_sp(2)];
+            % move (0,0) into center of the screen
+            gx_d = gx_d - width/2;
+            gy_d = gy_d - height/2;
+
+            %compute visual angle for gaze_x and gaze_y
+            s_x = gx_d; % x axis in spherical coordinates
```
eozd: These comments are wrong (`s_x`, `s_y`, `s_z`), or am I mistaken? This is a coordinate shift, not spherical coordinate transformation.

src/pspm_compute_visual_angle.m
```diff
@@ -93,92 +89,75 @@
...
+            s_y = distance * ones(numel(gx_d),1);% y axis in spherical coordinates, actually is the radial from participant to the screen
+            s_z = gy_d;  % z axis in spherical coordinates, actually is y axis on the screen
+            [azimuth, elevation, r]= cart2sph(s_x,s_y,s_z);% convert cartesian to spherical coordinates in radians, azimuth = longitude, elevation = latitude
```
eozd: I am confused why do we need this transformation? Can't we compute these angles in a more straightforward manner using `arctan` and `rad2deg`?

src/pspm_convert_pixel2unit.m
irojkov-ph: This is because the `chan.header.range` is a range of pixel which starts from 0, e.g. `[0,1023]` so there is in total 1024 pixels. Shall I specify it in a comment ?

src/pspm_convert_pixel2unit.m
irojkov-ph: Yeah sorry it was my previous solution and I've forgotten to change it.

src/pspm_compute_visual_angle.m
irojkov-ph: It will work I think, no ? Since `options.eyes` is an option specific to this function which can be 'right', 'left' or 'all', i.e. `options.eyes(1)` would then be respectively 'r', 'l' or 'a'. So the first two would work with `infos.source.eyesObserved`.
But I agree that this function do not follow the convention of `infos.source.eyesObserved`. Shall I change the convention for `options.eyes` ? (i.e. make it 'r', 'l' or 'lr')

src/pspm_compute_visual_angle.m
irojkov-ph: Copy that

src/pspm_compute_visual_angle.m
irojkov-ph: I would admit that I haven't checked these comments when I was modifying the function.
Indeed it is in cartesian coordinates and not in spherical yet. I'll change that.

src/pspm_compute_visual_angle.m
irojkov-ph: One can. Here is the algorithm, do you want me to do it ?
image

src/pspm_convert_pixel2unit.m
eozd: As far as I know, if the range is `[0, 1023]` then the pixel values are also in `[0, 1023]`. So an extra `+1` in the denominator would change the resulting ratio. For example, it can never be `1` anymore, meaning we would get slightly smaller results in the target unit.

src/pspm_compute_visual_angle.m
eozd: Ah now I see. I think having the same naming convention everywhere is a good idea, so yes please change it to 'l', 'r', 'lr'.

src/pspm_compute_visual_angle.m
eozd: I am a bit confused; what are the exact angles we need? If I understand correctly we need the following two angles:
1. the angle formed by
  1. the perpendicular line between eye and the center of the screen
  2. the line between eye and the horizontal projection of the gaze point against the center of the screen
2. the angle formed by
  1. the perpendicular line again
  2. the line between eye and the vertical projection of the gaze point against the center of the screen

Or do we need something else?

src/pspm_compute_visual_angle.m
irojkov-ph: Copy that

PR #111, opened 2020-05-04
Remove startsWith and endsWith from all functions

irojkov-ph:
Fixes a change request from @dominikbach. 

- In this pull request I have replaced all the occurrences of `startsWith` and `endsWith` from PsPM such that it is compatible with older than 2016 versions of Matlab.
- The test functions were also modified.

Please have a look at the changes and/or test it with your data before approving the pull request.
(I need at least one approval for merging)


sammaxwellxyz:
Ideally I'd test on 2015 or earlier but i do not yet have my UCL matlab license, and the free trial doesn't seem to permit getting older versions

src/backroom/set_blinks_saccades_to_nan.m
```diff
@@ -28,7 +28,7 @@
     %   - all elements in right data columns (except blink/saccade) that correspond to right
     %     blink or saccade rows are set to NaN
     if nargin == 3
-        fn_is_left = @(x) endsWith(x, '_l');
+        fn_is_left = @(x) strcmp(x(end-1:end), '_l');
```
sammaxwellxyz, line 31: Is there a reason for different endswith strategies?
this is index based, and then here for example is regex based
```
~isempty(regexp(msgline,'sec$'))
```
I'd assume the index is faster

irojkov-ph: Yes there is a reason, regexp allows to use a cell array and the index based method not. So I've used the latter one in situations when there was only one string and regexp when the variable could be an array of strings. I could have chosen regexp for all `endsWith` but it was easier to use the index method :sweat_smile: 

Do you prefer to have regexp everywhere ?

sammaxwellxyz: ok great, makes sense, no I think that what you've done helps to make the distinction

PR #116, opened 2020-05-27
Add condition to allow plotting any type of marker channel

### LINKED ISSUE #113: pspm_display shoud display any type of events signal
### Summary
Currently `pspm_display` allows to display only events signals of types 'markers' and 'hb' and all the other types are for now ignored by the function.

### Steps to Reproduce
1. Take a dataset which contains a marker channel
2. Rename by hand the marker channel 'chantype' variable with an other type (e.g. 'rs')
3.  Launch the function and choose to display the latter channel

### Expected Results
Display the events from the channel.

### Actual Results
The channel will not be displayed and a message saying "Nothing to display" will appear.

There is only two if conditions taking into account only 'markers' and 'hb' but no "else" condition.

### Technical Info
* PsPM revision ID: (latest)
* MATLAB version: 2019b

irojkov-ph:
Fixes #113 .

- Adds a new condition which allows the user to plot any type of marker channel and not only 'markers' or 'hb'. To do that I inspired myself from 'hb' for doing that.
-  It is a easy fix but in the future we should think of rewriting `ed_y_max_CreateFcn` in `pspm_display.m` because there is some redundancies, but for that it would need to be done with an experimentalists who would give advises on how to plot different signals. 
-  I let you check that pull request with your own data or with these ones: [test_data.zip](https://github.com/bachlab/PsPM/files/4687610/test_data.zip)


dominikbach:
All event channel types can be displayed now.
BUT only markers can be plotted in "integrated" mode - all other event types can only be plotted "extra". 
Desired functionality: all event channel types can be plotted as either extra or integrated.

irojkov-ph:
I've changed that, now all the marker channels are allowed to be plotted in 'extra' and in 'integrated' modes **except** for the 'hb' channel type since it was the initial behavior of this function (I thought that it shouldn't be changed). Tell me if I am wrong and if anything else is not correct.

irojkov-ph:

dominikbach:
No sure I'm looking at the right branch, but it is definitively different from the Master branch. 

What I see in bugfix/pspm_display now is:
- Markers can be plotted integrated or extra
- when I select a HB channel or any other channel, it is automatically plotted as extra (not integrated), and the extra/integrated option greys out after the channel has been plotted

So this is similar to my previous review.

irojkov-ph:
Thank you for the review.

I don't why it's the case for the other channels but 'hb' has indeed a particular treatment: it was originally implemented to be plotted as extra and not integrated (you can check that by checking out into the master branch and opening `pspm_display`).

So as far as understand you this was a wrong implementation, I'll then change that to have the possibility to select either `extra` or `integrated` mode for all types of events channels

irojkov-ph:
Now all the types of event channels can be plotted in `extra` or `integrated` mode, if someone can check that would be nice :)

src/pspm_display.m
```diff
@@ -765,7 +796,35 @@ function ed_y_max_CreateFcn(hObject, eventdata, handles)
...
+            if strcmp(handles.prop.event,'extra')
+                EVENTS(events,1)=min(wave)-.5;
+            elseif strcmp(handles.prop.event,'integrated')
+                temp=wave(events,1);
+                median_non_nan_vals = median(temp,'omitnan');
+                EVENTS(events,1)=temp;
+                EVENTS(isnan(EVENTS))=median_non_nan_vals;
+            end
+            
+            EVENTS(EVENTS==0) = NaN;
```
sammaxwellxyz, line 817: Hey, just checking i understand this bit

L806, EVENTS is set to a zero matrix
L808-L814, values are assigned to the EVENTS according to the plotting strategy and main data channel values, also setting NaN to the median value of the wave
L817 EVENTS has it's zero's set to NaN.

Does this mean there's an assumption that the wave values will never be 0, otherwise the events value would be overwritten to NaN?

```matlab
            events=round(events*sr.wave);

            EVENTS = nan(size(wave));

            if strcmp(handles.prop.event,'extra')
                EVENTS(events,1)=min(wave)-.5;
            elseif strcmp(handles.prop.event,'integrated')
                temp=wave(events,1);
                temp(isnan(temp)) = median(temp,'omitnan');
                EVENTS(events,1)=temp;
            end

```

Does this achieve the same without requiring that assumption?
1. initate EVENTS as NaN.
2. set EVENTS to the wave data after filling wave NaNs with median vals.
3. no need to set EVENTS = 0 to NaN

irojkov-ph: That's true I haven't noticed that, I've just inspired myself from the `marker` case, I'll check that

irojkov-ph:
I've implemented and tested your proposed changes @sammaxwellxyz , it works well, I let you review and if it's okay for you you can merge this PR ;)

sammaxwellxyz:
Hey @irojkov-ph , I have approved, however unable to merge as there's an outstanding request from @dominikbach

irojkov-ph:
Yep right, I just want to know his opinion on that PR since it's something that users tend to use quite often

PR #121, opened 2020-06-09
fix(butter): assign filt to variable from load as required by Matlab …

sammaxwellxyz:
…2020

Fixes (no issue)

- assign filt to variable from load in pspm_butter as required by Matlab 2020

### DIFF
--- a/src/pspm_butter.m
+++ b/src/pspm_butter.m
@@ -38,7 +38,7 @@ function [sts, b, a] = pspm_butter(order, freqratio, pass)
     if settings.signal
          [b, a]=butter(order, freqratio, pass);
     else
-        load('pspm_butter.mat');
+        filt = load('pspm_butter.mat').filt;
         switch pass
             case 'low'
                 f = filt{1};


irojkov-ph:
Looks good to me :+1:  Do you know if it is backward compatible, i.e. if it would work with e.g. Matlab 2010 ?

irojkov-ph:

sammaxwellxyz:
> Looks good to me 👍 Do you know if it is backward compatible, i.e. if it would work with e.g. Matlab 2010 ?

I do not, from matlab docs, this is the oldest I could find (2015a)...https://uk.mathworks.com/help/releases/R2015a/matlab/ref/load.html?searchHighlight=load
the oldest stackoverflow comment regarding load is 2011
https://stackoverflow.com/questions/4718417/matlab-loading-a-mat-file-why-is-it-a-struct-can-i-just-have-the-stored-vars

Do you know of a good way to find out? I think ideally we have CI get our back for us but that may take time to organise

sammaxwellxyz:
> @dominikbach @sammaxwellxyz I figured out today that the Jenkins server was down so I restarted it and now tests are running, after they finish we will merge some of the pull requests

Was there any success / failure with the tests? they don't seem to have been reported

irojkov-ph:
> Do you know of a good way to find out? I think ideally we have CI get our back for us but that may take time to organise

For checking the backward compatibility of a specific Matlab function you can just go to its documentation webpage and it's written on the very bottom when it was introduced (cf. picture).
image
The problem is that it only tells when it was introduced in the language but not when was the last time that its functionality changed. But I guess that this way of calling the load function: 
```filt = load('pspm_butter.mat').filt```
was there from the beginning it's just that I have never seen this before :smile:

irojkov-ph:
> Was there any success / failure with the tests? they don't seem to have been reported
All good ;) I've realized that Jenkins (the way the pipeline script is written) will not run tests on a PR if there are already some running on another one. Moreover this PR will not be put in a queue so after the current tests finish, no more tests will be ran.

This behavior is obviously not what we want so I will change it soon.

irojkov-ph:
You can merge the PR when the tests finish successfully ;)

Issue #118, opened 2020-06-01
pspm_data_editor should take missing epochs input


dominikbach:
### Feature Description
pspm_data_editor should take an existing missing epochs file as input, and display the epochs as if they had just been created, saving any changes to the same file.
Related issue: #117 (which would create such missing epochs files automatically)

PR #122, opened 2020-06-12
Feat/missing epochs file out

### LINKED ISSUE #117: pspm_simple_qa
### Feature Description

pspm_simple_qa should have an option of writing out missing epoch files, rather than setting detected artefact epochs to NaN directly. This will require an additional user-provided filename argument.

This will also require a change in pspm_pp (which calls pspm_simple_qa), and in the GUI.

Related issue: #118

sammaxwellxyz:
Fixes #117 #118  .

- Update artefacts job config to accept a artefacts epoch filepath to write to
- Construct epochs from qa filter and write to file if filename provided
- Update data editor to allow uploading an epoch file


irojkov-ph:

src/pspm_simple_qa.m
```diff
@@ -73,7 +73,33 @@
 filt = range_filter & slope_filter;
 d(filt) = data(filt);
 
+% write epochs to mat if missing_epochs_filename option is present
+if isfield(options, 'missing_epochs_filename')
```
irojkov-ph, line 77: Can you add something like ` && ischar(options.missing_epochs_filename)` ? And if it's not can you send a warning please ? Usually we do not trust the user :sweat_smile:

irojkov-ph:
I've just tested with my data. Everything looks good to me except few details:
- Can you please write a description of the new option in the header of `pspm_simple_qa` ?
- In the GUI its not very clear why there is a 0 when the option "not saving" is selected, it seems that one has to change it to 1  in order to save missing epochs, can you change it to an non-interactive field ? (I have no idea if it's possible, I'm not very familiar with gui)
- Can you please specify in gui and in the description that the filename has to be without any extension otherwise the file will not be readable at the end ? (I've checked it by setting 'test.txt' as a filename)

sammaxwellxyz:
> I've just tested with my data. Everything looks good to me except few details:
> 
> * Can you please write a description of the new option in the header of `pspm_simple_qa` ?
> * In the GUI its not very clear why there is a 0 when the option "not saving" is selected, it seems that one has to change it to 1  in order to save missing epochs, can you change it to an non-interactive field ? (I have no idea if it's possible, I'm not very familiar with gui)
> * Can you please specify in gui and in the description that the filename has to be without any extension otherwise the file will not be readable at the end ? (I've checked it by setting 'test.txt' as a filename)

great, thanks for the feedback will take a look

irojkov-ph:
Is this a work in progress or is it finished ? I see that you pushed some changes since my approval so I am a bit confused :sweat_smile:

sammaxwellxyz:
Hey @irojkov-ph , thanks for reviewing, Dominik had emailed me about some changes to make which I implemented regarding using the data editor function directly. There are some further points to address as Dominik had mentioned in slack about the actual simple qa implementation. @dominikbach how do you feel about separating them out and having this pr as a mostly GUI piece and a separate issue and PR as the actual simple qa implementation?

PR #131, opened 2020-06-30
Add an option to take into account random ITIs

irojkov-ph:
Fixes an issue spotted by Jelena.

Issue description:
When trying to split session per blocks using `pspm_split_sessions` in data where marker distances are not constant (random ITI), you can end up with session which contains more markers than needed. 
Example: you have 40 markers in data and 4 blocks, you want to split uniformly, but due to random marker distances (not going much in details) the session 3 is found to have 11 markers instead of 10 such that after processing and recombining the data you will have 41 markers instead of 40.

- In this pull request, I implemented an option which I called `randomITI` and which can take the values 1 or 0 (default: 0)
- What it does is to evaluate the mean marker distance not per session but on the all the markers.

I am waiting for Jelena to confirm me that it reproduces the results correctly (for data which worked well) and then I will submit for review.

### DIFF
--- a/src/pspm_split_sessions.m
+++ b/src/pspm_split_sessions.m
@@ -30,6 +30,10 @@ function newdatafile = pspm_split_sessions(datafile, markerchannel, options)
 %                           Default = 0
 % options.verbose           Tell the function to display information
 %                           about the state of processing. Default = 0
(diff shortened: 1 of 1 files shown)


irojkov-ph:
All good, ready to be reviewed

PR #136, opened 2020-07-22
import_viewpoint doesn't import the right number of datapoints

irojkov-ph:
Fixes a bug that I've discovered while importing a viewpoint file. Here is a test file (anonymized): [test_data.txt](https://github.com/bachlab/PsPM/files/4959048/test_data.txt)

*Bug description:*
For importing a viewpoint file in PsPM, `import_viewpoint` function uses `textscan` Matlab function to read formatted data from the text file. The format is determined by a `formatSpecs` character array which is created by `get_columns_to_read` function (cf. `import_viewpoint.m` file), e.g. `%f\t%f\t%f\t%f\t%f\t%f\t%f\t%f\t%f\t%f\t%f\t%s` (num. of col depend on the file). But as one can see it already contains the delimiter and thus it shouldn't be specified in options of the `textscan` function. Moreover when specifying it the `textscan` stops at the first marker because `formatSpecs` do not match anymore.

*Reproducing the bug:*
Execute: `import_viewpoint('test_data.txt')`. You'll see that the data array contains only 15k points instead of the 200k and in the marker field there is only one point (starting marker '+,+').

*Expected result:*
Execute: `import_viewpoint('test_data.txt')` (after making the changes). You'll see that the data array contains the right number of points and in the marker field there are all the specified point.

*Changes proposed in this pull request:*
- Remove 'Delimiter' option in `textscan` function in the `parse_viewpoint_file` function used by `import_viewpoint`

### DIFF
--- a/src/Import/viewpoint/import_viewpoint.m
+++ b/src/Import/viewpoint/import_viewpoint.m
@@ -114,7 +114,7 @@ function [dataraw, marker, messages, chan_info, file_info] = parse_viewpoint_fil
         begidx = linefeeds(msg_line) + 1;
         str(begidx : begidx + 1) = '/';
     end
-    C = textscan(str, fmt_str, 'Delimiter', '\t', 'CollectOutput', 1, 'CommentStyle', '//');
+    C = textscan(str, fmt_str, 'CollectOutput', 1, 'CommentStyle', '//');
     dataraw = C{1};
     marker = C{2};
 
--- a/test/import_viewpoint_test.m
+++ b/test/import_viewpoint_test.m
@@ -83,7 +83,7 @@ classdef import_viewpoint_test < matlab.unittest.TestCase
             % ---------------------------------------------------------------------------
             msg_counter = 1;
             for line = datalines
-                parts = split(line, sprintf('\t'));
+                parts = strsplit(line{1},'\t');
                 msg = parts{marker_index};
                 if ~isempty(msg)
                     tbeg = to_num(parts{2});


sammaxwellxyz:
@irojkov-ph could you update the import_viewpoint test / data so that it would have failed before the fix and now passes?

irojkov-ph:
> @irojkov-ph could you update the import_viewpoint test / data so that it would have failed before the fix and now passes?

Sorry I don't get what you mean ? :sweat_smile:

irojkov-ph:
Okay I've looked at the test data and this is what I've found:
If the data are in the format:
```
10	554.0725	  16.7128	0.4126	0.9307	-1	0.0482	0.0172	1	0.0167	33215		
10	554.0891	  16.6496	0.3320	0.9216	-1	0.1104	0.0396	1	0.0166	33216   CS1
10	554.1058	  16.7102	0.3380	0.9342	-1	0.1658	0.0207	1	0.0334	33217		
```
Then the file is imported correctly til its end (i.e. no issue). If on the otherhand the file is in the following format (notice the double tabulation before the marker):
```
10	554.0725	  16.7128	0.4126	0.9307	-1	0.0482	0.0172	1	0.0167	33215		
10	554.0891	  16.6496	0.3320	0.9216	-1	0.1104	0.0396	1	0.0166	33216       CS1
10	554.1058	  16.7102	0.3380	0.9342	-1	0.1658	0.0207	1	0.0334	33217		
```
Then the `formatSpecs` do not match anymore when the function `textscan` reaches this marker.
Both cases work well with the proposed changes.

@sammaxwellxyz : May I ask you to review it once again ? Or give me your opinion ?

@dominikbach : what should I do ? Can we merge this modification ? Shall I still open an issue ? What was the content of it already ?

irojkov-ph:
I've modified the existing test file such that it reproduces this error. I let the checks to finish and @sammaxwellxyz to review in order to merge this pull request. I'll create an issue for checking the import functions.

PR #146, opened 2020-08-20
feat(filter-epochs): include slope threshold and use array ops instea…

### LINKED ISSUE #128: pspm_simple_qa: epoch output path
### Feature Description

Use full path for saving epoch output file, don't change file extension (i.e. if no file extension then add 'mat' and if there is one then keep it - this is the behaviour of matlab's 'save') and ask for path in GUI (similar to model file user interaction for pspm_glm)

### LINKED ISSUE #129: pspm_simple_qa improvements
### Feature Description

Separate artefacts from small oscillations when recordings are made with low-precision settings, leading to small oscillations between adjacent discrete values, with large slope.

1. separate upward from downward deflections (i.e. mulitplying
slope_filter with sign(diff(data)))

2. then find contiguous epochs with same sign in the slope filter

3. exclude this epoch from the filter if the overall deflection over this epoch is below
an amplitude threshold (e.g. 0.05).

4. Find and exclude 'islands' of data that are flanked by missing epochs, and are shorter than a duration threshold
(e.g. 1 s). One solution would be the image processing toolbox but perhaps there is a simple way with core matlab functions.

sammaxwellxyz:
Fixes #129  .

- add a threshold for the slope filter
- add data island filter

Requires:
 - testing, I'd suggest creating a file with defined data so that the epochs that are filtered can easily be tested for the various scenarios of oscillations / data islands etc and the options that go into the function. But, if there is good data available to use and confidently get expected results then that would also suit


irojkov-ph:
Sam's handover comment:

> The epoch PR should fix issue 129, I've tested the functionality of individual chunks of the code, such as using array operations to get epochs; filtering data islands using a time threshold; and the slope deflection amount threshold. However, there are no automated tests that actually test the correct epochs are extracted, there are tests that run the function but not assert the output is as expected. I've held off propagating the help texts through to the GUI and docs as these may be changing with the upcoming release and I expect this piece to require more testing before releasing (or at least saying it's released). Dominik, I'm not sure it meets the exact approach discussed. For oscillation filtering, the code should find the range between all values in an epoch, if that range exceeds a set threshold the epoch can be filtered. This way if the threshold is set to the oscillation voltage jump, then oscillations should not be considered filtered epochs.

irojkov-ph:
This pull request fixes #128 and #129, i.e. it adds the output directory for the artefact epochs file as well as all the options from `pspm_simple_qa` in the GUI. 

@dominikbach : before merging may I ask you to test the 'Artefact removal' functionality using the GUI please ? It seems to work but I don't have any sample file nor the exepected result.

dominikbach (approved):
I've tested the GUI with DoxMemP data. All options seem to work fine.

However, I realised the following defaults don't work together:
deflection_threshold = 0
expand_epochs = 0.5
because for data that oscillates between discrete values, where many individual data points are detected as artefacts and then expanded by 1 s, no data will remain, and this could confuse the user.
So for convenience, can we set the default deflection threshold to 0.1 in function and GUI, and then this will work fine.

PR #151, opened 2020-08-25
Feat/dsv 110

### LINKED ISSUE #110: Add support for Delimiter Separated Values data files
### Feature Description
The goal is to be able to import **Delimiter Separated Values**, particularly .csv and .tsv

### How
Modify .txt import so as to allow for delimiters that are non matlab standard (white space, backspace, tabulator). One way would be to use the excell switch (which we have to handle anyways): "sep=<delimiter>" in the first line of the file. But this is non standard dsv. I'll implement this as a quick fix for #109 

In addition, we might want to switch on the file extension and assume delimiters based on this. 
This raises the issue that there does not seem to be an accepted standard. while .tsv is not unheard of, .ssv is quite obscure and most semi-colon dsv are saved as .csv, meaning sanity checks would have to be implemented. 

| extension | delimiter |
|-|-|
|.csv|, (coma)|
|.tsv|\t (tab)|
|.psv|\| (pipe)|
|.ssv|; (semi-colon)|
 
An other option would be to pass the separator as an argument to the function. this would mean that GUI and parent functions would have to be adapted

### Questions
- which of these options should be implemented?
- would we replace the text import function or create a new import?

sammaxwellxyz:
Fixes #110  .

I  have since noticed there is another branch open that looks like it seeks to achieve the same goal.
`Feature110/dsv_import`

- DSV import options
- CSV import option

Based on the CSV implementation just being a helper of the DSV it should be simple to extend for other convenience separators, e.g. pipes or semi-colons


src/pspm_cfg/pspm_cfg_import.m
```diff
@@ -176,6 +176,12 @@
 smi_stimulus_resolution.help      = {['The resolution of the stimulus window. This field is required' ...
                                       'to perform px to mm conversions for gaze channels']};
 
+delimiter           = cfg_entry;
+delimiter.name      = 'Delimiter';
+delimiter.tag       = 'delimiter';
+delimiter.strtype   = 's';
+delimiter.help      = {['The delimiter to be used for file reading, leave blank to ']};
```
irojkov-ph, line 183: I think you forget to finish the sentence xD

src/pspm_get_txt.m
```diff
@@ -1,10 +1,11 @@
-function [sts, import, sourceinfo] = pspm_get_txt(datafile, import)
+function [sts, import, sourceinfo] = pspm_get_txt(datafile, import, delimiter)
```
irojkov-ph, line 1: Again as I pointed it out last time, I prefer to be consistent in functions names and arguments, for different data file type there should be only `datafile` and `import` as argument (look at `pspm_get_eyelink`, `pspm_get_physlog`, etc.) So I propose to put this `delimiter` option into the `import` structure, or something like that.

src/pspm_import.m
```diff
@@ -32,7 +32,9 @@
 %              Can be one of the following units:'mm', 'cm', 'm','inches'.
 %           - .denoise: for marker channels in CED spike format (recorded 
 %              as 'level'), filters out markers duration longer than the 
-%              value given here (in ms) 
+%              value given here (in ms)
+%           - .delimiter: for delimiter separated values, value used as delimiter for file read
+%
```
irojkov-ph, line 37: If in `pspm_import` you put the `delimiter` in the `import` structure, why in `pspm_get_txt` you separate it from the structure ? (comment related to the comment in `pspm_get_txt`)

test/pspm_get_superclass.m
```diff
@@ -47,7 +47,12 @@ function valid_datafile(this)
             fprintf('\n');
             
             for k = 1:numel(this.testcases)
-                [sts, import, sourceinfo] = this.fhandle(this.testcases{k}.pth, this.testcases{k}.import);
+
+                if (isfield(this.testcases{k}, 'extra_args'))
+                    [sts, import, sourceinfo] = this.fhandle(this.testcases{k}.pth, this.testcases{k}.import,this.testcases{k}.extra_args{1});
+                else;
+                    [sts, import, sourceinfo] = this.fhandle(this.testcases{k}.pth, this.testcases{k}.import);
+                end;
```
irojkov-ph, line 55: What does this do ?

test/pspm_get_txt_test.m
```diff
@@ -45,6 +45,48 @@ function define_testcases(this)
...
+            %generate testdata
+            header = {'scr' 'ecg' 'heart' 'resp'};
+            data = rand(900, 4);
+            
+            fid = fopen(this.testcases{4}.pth, 'w');
+            fprintf(fid,'scr|ecg|rate|resp\n');
+            for k=1:size(data,1)
+                fprintf(fid,'%f|%f|%f|%f\n', data(k,1), data(k,2), data(k,3), data(k,4));
+            end
+            fclose(fid);
```
irojkov-ph, line 89: Nice :+1:

irojkov-ph (changes requested):
In general good, I haven't tested it with the files you've specified but looks good :+1: !
Just few more comments:
- the testing files should be deleted from the folder (I guess you know)
- could you also add these changes in the Manual and the `release_notes.tex`, I think it could be nice to already publish it in the next release :smiley:

src/pspm_cfg/pspm_cfg_import.m
sammaxwellxyz: ha ha thanks

src/pspm_get_txt.m
sammaxwellxyz: I remember being conflicted on putting it in import for some reason but I'll take another look, I was likely just a bit confused

test/pspm_get_superclass.m
sammaxwellxyz: This is because I changed the function signature of the pspm_get_text so I needed the superclass to be able to handle extra args. But should be able to remove once I make the pspm_get_text the same as the others again

src/pspm_get_txt.m
sammaxwellxyz: Ah yes, so the `import` seems to me to be associated with a specific channel, not the file. Is that right? It would mean something like adding the delimiter to all import cells, then just taking the first and considering it gospel. Shall I do that? It's not particularly nice but I think that's the best way without an extra arg

sammaxwellxyz: Or just put it on the first cell in the import cell array, it's hacky either way but just on first I guess is a smidgen less memory

irojkov-ph: Yes that's correct, I think we can do it like for the `eyelink_trackdist` in the `pspm_get_eyelink` function where we are including it in every channel, it's a bit redundant but it's kind of logical, so what you can do is to include the delimiter in every channel of the `import` structure

sammaxwellxyz:
@irojkov-ph thanks! all points should now be addressed.
- Removed unnecessary test files (they're generated in tests)
- added release notes
- used delimiter on import structure
- made pspm_get_csv a separate function instead of partially application

src/pspm_get_csv.m
```diff
@@ -0,0 +1,19 @@
+function [sts, import, sourceinfo] = pspm_get_csv(datafile, import)
```
irojkov-ph, line 1: Great :+1:

src/pspm_import.m
irojkov-ph: :+1:

irojkov-ph (commented):
Looks good to me :ok_hand:

irojkov-ph (approved):

PR #153, opened 2020-08-26
Bugfix/triming indices

irojkov-ph:
Fixes a bug spotted by Jelena.
She was trimming data but then while analyzing them PsPM told her that few files have a wrong structure, mainly that they are out of the range `[0, infos.duration]`. I investigated that and I concluded that there is a problem in the definition of the starting point index: it is defined as if the indices starts from 0 and not from 1 as it is in Matlab; secondly we defined the starting and ending indices with `floor` function that means if the trimming point do not correspond exactly to a data point there is a potential risk to take an additional point which time would be outside the trimming window.

It doesn't impact much all the previous analysis made with PsPM since at maximum we were ignoring 2 point (negligible over many points).

- Add small changes in gitignore
- Replace `floor` by `ceil` for the starting point index and add "+1" according to Matlab convention
- Add changes in the Manual and the release notes for the release of next week


dominikbach (approved):
Looks good to me.

PR #161, opened 2020-09-16
Bugfix in pspm_dcm related to minITI 

irojkov-ph:
Fixes a bug spotted by Karita.

Merge the #160 first before merging this pull request !

- On line 548: Make sure that the trials which are enabled and have the 'current' subsession id are more than one for calculating the minITI according to 63f408d
- On lines 655 and 703: Change one argument which is sent to the function `get_data_after_trial_filling_with_nans_when_necessary`, mainly replace `sbs_iti` by `model.iti` since all the rest of the variables are redefined on lines 585-594 so logically using `sbs_iti` is wrong I would say.


dominikbach (approved):
Sounds reasonable.

PR #168, opened 2020-11-23
Last trial filtering

teddychao:
- Added an option options.lasttrl to select desired trials and remove insufficient trials. Default as 7.
- Added an error variable to record the filtered insufficient trials.
- Added the feature to preserve the satisfactory trials and abandon the insufficient trials.


teddychao:
I am not sure if it is enough to update only _sbs_data_. I noticed the variables related to the last ITI were updated between line 585 and 594, thus I updated the _sbs_data_ before that code period. However, I am not quite sure if there are other variables that shall be updated too..

src/pspm_dcm.m
```diff
@@ -582,6 +583,26 @@
...
+    % Find the position of the target trial in proc_subsessions
+    % Check the interval since the start of the last trial
+    error_log(i_session)=sum(sbs_iti{i_session}<options.lasttrl);
+    if error_log(i_session) > 0
+        i_trial = length(sbs_iti{i_session});
+        last_trl_start = sbs_trlstart{i_session};
+        last_trl_start = last_trl_start(i_trial)*model.sr;
+        last_trl_stop = numel(sbs_data{i_session,1});
+        sbs_data_out{i_session,1}(last_trl_start:last_trl_stop) = [];
+    end
```
irojkov-ph, line 602: In other words: why are you not using in this "if" statement the information about which ITI is too short ?

src/pspm_dcm.m
```diff
@@ -234,6 +234,7 @@
 try options.crfupdate; catch, options.crfupdate = 0; end
 try options.eventnames; catch, options.eventnames = {}; end
 try options.trlnames; catch, options.trlnames = {}; end
+try options.lasttrl; catch, options.lasttrl = 7; end
```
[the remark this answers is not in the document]
irojkov-ph: Also: Why your parameter is called last trial ?

irojkov-ph (changes requested):
In general, the implementation is good but I don't understand the reasoning behind it, you want to remove insufficient trials but you remove only *the last* trial if *any* other trial in the session is too short.
@dominikbach : Is my concern justified ? Or am I completely mistaken ?

src/pspm_dcm.m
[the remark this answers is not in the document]
teddychao: Thank you for the suggestion. I have added the description on the latest version. The name was set to be last trial because it is used to filter the last trials with insufficient information.

src/pspm_dcm.m
```diff
@@ -582,6 +583,26 @@
...
+% Remove the trial that has insufficient information
+sbs_data_out = sbs_data;
+flag_valid = ~cellfun(@isempty, sbs_trlstart);
+error_log = zeros(size(sbs_iti));
+% Do processing in the index of valid sessions
+idx_session = nonzeros((1:size(sbs_data,1)).*flag_valid);
+for i_session = idx_session'
+    % Find the position of the target trial in proc_subsessions
+    % Check the interval since the start of the last trial
+    error_log(i_session)=sum(sbs_iti{i_session}<options.lasttrl);
```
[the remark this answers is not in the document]
teddychao: Apologies, this is due to my misunderstanding of the variable. I thought the insufficient information was happening for only the last trials. I have corrected this mistake in the updated code. Thanks.

src/pspm_dcm.m
teddychao: Sorry I am not quite clear about this question. Instead, I can explain the idea here. The idea is we try to find the sessions where the last trial is with insufficient information and remove the last trial in such sessions. Such sessions with this issue will be recorded into the variable _error_log_.

irojkov-ph:
I don't see your changes @teddychao , have you pushed them ?

src/pspm_dcm.m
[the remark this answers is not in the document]
irojkov-ph: Okay makes sense now ! Happy to hear that my concern was helpful to spot a bug :sweat_smile:

src/pspm_dcm.m
irojkov-ph: Do you store this variable in the final model structure ?

irojkov-ph (changes requested):
See my previous [comment](https://github.com/bachlab/PsPM/pull/168#issuecomment-732449208)

src/pspm_dcm.m
teddychao: Yes, I have updated the code to add storing this variable in the final model structure as _dcm.lasttrialfilter_.

src/pspm_dcm.m
```diff
@@ -809,6 +843,7 @@
 dcm.modeltype = 'dcm';
 dcm.modality = settings.modalities.dcm;
 dcm.revision = rev;
+dcm.lasttrialfilter = error_log; % recorded the sessions that have last trial filtered
```
irojkov-ph, line 846: I suggest to add it in the `model` structure (somewhere btw lines 623-633) because I find it natural to have the information about excluded trials in the input model for the inversion process (at the end `lasttrialfilter` will be stored in `dcm.input` as well as in `dcm.invmodel`).

irojkov-ph (approved):
All good, except the last small comment, so I changed it and approved the PR, you can merge it whenever you are ok with what I've done. Thanks @teddychao ! :+1:

src/pspm_dcm.m
```diff
@@ -77,6 +77,12 @@
 % - options.eventnames: cell array of names for individual events, in the
 %   order they are specified in the model.timing array - to be used for
 %   display and export only
+%
+% filtering options
+% - options.trialfilter: a cut-off value set for the interval since the start
+%   of the last trial. Default as 7s. The last trials with insufficient
+%   information (less than 7s) will be removed. Can be set as infinity to
+%   ignore this filtering.
```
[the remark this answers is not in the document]
dominikbach: You are completely right - 7 s is a bit arbitrary. However, there were previous questions from users that had encountered implausible parameter values with almost no data after the last trial; so for the more naive user I think some kind of threshold is required here. 

This figure from Bach et al. 2010 shows that after 7 seconds, average SCR elicited by various types of events have peaked and decayed to about 80% of the peak value, which heuristically gives a reasonable chance of estimating the peak amplitude. But we could just as well motivate 10 s (60%) or 15 s (40%).
image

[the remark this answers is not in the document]
irojkov-ph: Fair enough ! I agree with 7s then :wink:

src/pspm_dcm.m
irojkov-ph: Changed it myself

teddychao: I agree with you as it should be known in the input model.

dominikbach:
DCM is probably too time-consuming to be included in the automatic tests. I'd suggest two manual tests to ensure its functionality at least at the current stage. First, a test including one participant from the DCM tutorial data set; secondly, an altered data file from this set with less than 7 seconds after the last trial. 

https://github.com/bachlab/PsPM-tutorial-datasets/releases/download/tutorial-datasets/Tutorial_dataset_DCM.zip

teddychao:
[Last_trial_filtering.pdf](https://github.com/bachlab/PsPM/files/5840142/Last_trial_filtering.pdf)
@dominikbach Could I ask you to check the results of the last trial testing based on the provided dataset? Thanks.

teddychao (commented):
Convert into integers for selecting the elements at the end of the array.

src/pspm_dcm.m
```diff
@@ -602,7 +602,7 @@
         % Find the position of the target trial in proc_subsessions
         last_trl_start = sbs_trlstart{i_session};
         % Convert from time (s) to data points
-        last_trl_start = last_trl_start(i_trial)*model.sr;
+        last_trl_start = ceil(last_trl_start(i_trial)*model.sr);
```
teddychao, line 605: Convert into integers for selecting the elements at the end of the array.

dominikbach: @teddychao can we discuss the posted document in our next meeting? Also, can you just confirm that no trials are removed when the data available after the last trial exceeds the threshold? Thanks!

teddychao:
Updates:

1. options.lsttrlcut allows users to specify a cut-off value to filter last trials.
2. options.lsttrlkeep allows users to choose whether to remove the insufficient last trials or preserve them in the outputs.
3. options.lsttrlpath allows users to save the time interval of the filtered last trials into a .mat file.
4. Users will be warned if options.lsttrlpath is set to empty (not to save the filtering results) and options.lsttrlkeep is set to 0 (remove the insufficient last trials)

teddychao:
Update 24 Jan 2020

- The updates of adding data removal in the previous updates have been removed in this version.
- The dcm.stats was updated for showing the results of filtering. Specifically, the results of dcm.stats for the last trial that does not contain sufficient information will be set to NaN.

dominikbach:
The code appears to work now. I made a few more notes on Slack, regarding the naming of the option, its place in the function header, and the matlabbatch GUI.

teddychao:
> The code appears to work now. I made a few more notes on Slack, regarding the naming of the option, its place in the function header, and the matlabbatch GUI.

Thank you Dom, could you please kindly add me (dadi.zhao@ucl.ac.uk) to the slack group so I can check your comments? Many thanks

Best wishes,
Ted

teddychao:
Under the data option, the last trial filtering has been available now

teddychao:
The typo mistake has been fixed now
Screenshot 2021-02-01 at 01 28 58

PR #170, opened 2020-11-26
Issue 158

### LINKED ISSUE #158: Allow pspm_simple_qa to just write artefact epochs and leave data unchanged
### Feature Description

- add new option for this, which can be called from the matlabbatch through artefact removal
- if desired return original data but do write artefact epochs to file

teddychao:
Fixes #158.
Allow users to save artefact epoches.
Allow data to be unchanged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,5 @@ vendor/
 *.DS_Store
 
 # Ignore MATLAB sync files
-.MATLABDriveTag
-.asv
+*.asv
+*.MATLABDriveTag
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,10 @@ vendor/
 # Ignore all the temporary files
 *.lyx~
 *.m~
+
+# Ignore all macOS files
+*.DS_Store
+
+# Ignore MATLAB sync files
+.MATLABDriveTag
+.github/ISSUE_TEMPLATE/.MATLABDriveTag
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ vendor/
 # Ignore MATLAB sync files
 *.asv
 *.MATLABDriveTag
+*.mat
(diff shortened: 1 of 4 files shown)


dominikbach (changes requested):
Great start, a few comments:
- this is not clear from the formulation of the issue, but I wonder now: if change_data is set to false, would it not be more useful not to write a new data file (as it will just contain the same information)
- for legacy reasons, PsPM uses numerical variables with values "0" and "1" instead of booleans. 
- why can we not use a missing epochs file name with extension? The extension stated by the user can just be dropped and replaced with "mat".

teddychao:
> this is not clear from the formulation of the issue, but I wonder now: if change_data is set to false, would it not be more useful not to write a new data file (as it will just contain the same information)
I agree with this comment. I have added the warnings reminding users that they have set duplicated outputs or insufficient data saving in the options. Users might still insist to store the artefact epochs and the changed data if they prefer to ignore the warnings.
> For legacy reasons, PsPM uses numerical variables with values "0" and "1" instead of booleans.
Thanks for reminding, I have corrected this part of the code.
> Why can we not use a missing epochs file name with extension? The extension stated by the user can just be dropped and replaced with "mat".
I agree. The current code is able to allow users to set the filenames only instead of the combination of the filename and the extension.

dominikbach:
(1) After going through the code, I see that there is a duplication: options.missing_epochs_filename and options.artefact_epochs_filename do the same thing (the first one is implemented more robustly, so the second option can be removed)

(2) When I set "change_data" to 0, I still obtain a new data file prepended with "m" that contains the changed data. There should be no file output in this case.

(3) It's good that the user can select a file name without extension; but there should be no error if the user does chose an extension.

dominikbach:
Here are some test data.

[Uploading DoxMemP_wdq_02_sn1.mat.zip…]()

dominikbach (approved):
(1) ok
(2) is due to the call through pspm_pp, which may not be a good idea - I'll open a separate issue for this.
(3) was not fixed - I'll remove the error in ll. 110-113

PR #174, opened 2021-01-07
Fix issue 133

teddychao:
- Make pspm_spit_sessions to use pspm_trim.
- pspm_trim can be updated to improve the compatibility.


teddychao:
This branch can be updated after fixing issue 196 as the loop of asking users for file overwriting often appears in the current develop branch.

teddychao:
Could be updated after the approval of the latest version of _split_sessions_

dominikbach:
There are typos in pspm_split_sessions ("makerinfo" instead of "markerinfo") that will be fixed when the function calls pspm_trim (where the typo was fixed in an earlier update).

teddychao:
I think this version has been sufficient for the next release. However, I think there is lots of potential code optimisation that is helpful for code maintenance, although I have done some in this version. This version can be used for previewing and ready for the release as it is functioning as expected now.

dominikbach (approved):

PR #179, opened 2021-01-18
Issue 173

teddychao:
Make pspm_simple_qa independent from pspm_pp
(1) Rename pspm_simple_qa into pspm_scr_pp
(2) Call pspm_scr_pp directly rather than through pspm_pp (requires adding file I/O)
(3) Change Matlabbatch GUI to reflect this
(4) If change_data = 0, do not write a new data file


teddychao:
The simple_qa is out of date in this version. I will open a new pull request from the develop branch.

PR #189, opened 2021-02-09
Issue 188

teddychao:
Improves the output of simple_qa.
The epochs output will include both missing epochs and detected clipping.
The old clipping output has been removed in this version.


teddychao:
Merged into issue 186

PR #190, opened 2021-02-10
Correct minor error in comment in pspm_simple_qa

dominikbach:
Minor issue.


teddychao (approved):

src/pspm_simple_qa.m
```diff
@@ -183,7 +183,7 @@
 if isfield(options, 'missing_epochs_filename')
     if ~isempty(find(filt == 0, 1))
         epochs = filter_to_epochs(filt);
-		epochs = epochs / sr; %convert into integers
+		epochs = epochs / sr; %convert into seconds
```
teddychao, line 186: Sorry I added wrong comments... I was thinking about converting it into integers and then reminded we shall use the direct output without rounding, but I forgot to change the annotation, my fault..

PR #193, opened 2021-02-19
Fix issue 192

teddychao:
- Added warning to let users know a potential reason why no results are generated.
- Improved function menu into question dialog to set default values for a question dialog.
- Default fallback of overwriting dialog is set to 'Yes'

### DIFF
--- a/src/pspm_dcm.m
+++ b/src/pspm_dcm.m
@@ -272,9 +272,9 @@ end
 
 % check files --
 if exist(model.modelfile, 'file') && options.overwrite == 0
-    overwrite=menu(sprintf('Model file (%s) already exists. Overwrite?', ...
-        model.modelfile), 'yes', 'no');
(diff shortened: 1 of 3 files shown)


teddychao:
@dominikbach I tried to make the code throw a warning if the max delay is set to be very small. However, it looks that the test data on Jenkins will still have a very small max delay, and no sound will be detected then. I got some testing data from Jelena, and that data has a max delay as 1.8s. In that case, I cannot find any sound detected as no sound satisfy this limit. Could we discuss about this later? Thanks.

teddychao:
@dominikbach I have updated pspm_find_sound by implementing Samuel's solution to the issue of handling two sounds. However, the warning of checking if the maxdelay is too strict cannot be added. It looks like the maxdelay on Jenkins testing files has been very small.

dominikbach (approved):

PR #194, opened 2021-02-21
Fix issue 173

teddychao:
1. Rename pspm_simple_qa into pspm_scr_pp
2. Call pspm_scr_pp directly rather than through pspm_pp (requires adding file I/O)
3. Change Matlabbatch GUI to reflect this
4. If change_data = 0, do not write a new data file


teddychao:
GUI information for _pspm_scr_pp_ has been updated, but the help text needs updates. Could be discussed later with @dominikbach .

dominikbach (changes requested):
See Slack for requested changes.

teddychao:
_Update 22nd March_
The _scr_pp_ will be tested by using a separated test function _pspm_scr_pp_test_.

teddychao:
I think this version has been sufficient for the next release. However, I think there is lots of potential code optimisation that is helpful for code maintenance, although I have done some in this version. This version can be used for previewing and ready for the release now.

dominikbach (changes requested):
pspm_scr_pp should be a standalone function, including file i/o.

teddychao:
> pspm_scr_pp should be a standalone function, including file i/o.

Sorry Dom I am a little confused about the output. I guess the expected output would be in the same format as the input. I do not see any question if the input is a struct. However, if the input is a file, and what would be the output? This is because the input would be a filename string in this case, and this filename string should not be changed after processing. I think this issue happens in _split_sessions_ too, but I do not see any code used _split_sessions_ with a filename, so currently there is no issue for _split_sessions_...

dominikbach:
This is a bit inconsistent across PsPM functions, unfortunately.

In this case, the function should (depending on the input)
- write a new data channel with artefacts periods replaced with NaN
- output an epochs file with the artefact periods
- or both

I guess a useful output would be [sts, outfile(s)] where outfile(s) are the data file (unchanged from the input) and/or the epochs file?

dominikbach:
As discussed. Input options:
- write modified data or not (i.e. artefact periods replaced with NaN values)
- write epochs file or not
- both
- in case of write modified data, additional option "channel_action" as in pspm_emg_pp

teddychao:
Output:
 - original file name if .epochfilename is not specified
 - epoch file name if .epochfilename is specified

dominikbach (changes requested):
- pspm_cfg_run_artefact_rm.m  seems to load the data and then run pspm_scr_pp with a struct - that should just be a file name?
- the input argument "sr" in pspm_scr_pp is not needed?
- if you change the default behaviour of channel_action, then this must be changed in all preprocessing functions (not just pspm_emg_pp) and presumably in a separate pull request, also it needs to be documented in the manual. For now I suggest to keep the default as replace and open an issue that this could/should be changed and why.

teddychao:
> * pspm_cfg_run_artefact_rm.m  seems to load the data and then run pspm_scr_pp with a struct - that should just be a file name?

I agree. The * pspm_cfg_run_artefact_rm.m* seems to be using the old * pspm_scr_pp* where the file loading has not been updated. It is currently updated to load data file directly instead of reading struct from file.

> * the input argument "sr" in pspm_scr_pp is not needed?

I agree, the sample rate input seems redundant and should be loaded from the data file header. I have updated *pspm_scr_pp* to do so. However, it seems that *sr* is used by various functions as the input. Would u recommend to generalise *sr* by proposing a new pull request?

> * if you change the default behaviour of channel_action, then this must be changed in all preprocessing functions (not just pspm_emg_pp) and presumably in a separate pull request, also it needs to be documented in the manual. For now I suggest to keep the default as replace and open an issue that this could/should be changed and why.

I agree. The channel action is currently using _replace_ as the default option. A new issue has been proposed [here](https://github.com/bachlab/PsPM/issues/237).

dominikbach (changes requested):
This looks almost finished.

Some remaining comments.

1. GUI options and helptexts should be copied from the "preprocessing SCR" option in artefact removal.
2. GUI for artefact removal and the associated function pspm_pp should have no option to call "preprocessing SCR" or pspm_scr_pp (there should be only one way of calling pspm_scr_pp).
3. There is a warning on PsPM startup: "Item 'Preprocessing SCR', field 'prog': Value must be a function or function handle on MATLAB path."

teddychao:
Looks like _preprocessing SCR_ is missing in the main GUI but available in batch editor, so I will add it...

teddychao:
@dominikbach not sure if *scr_pp* or *pp_scr* is preferred. I think EMG data preprocessing is currently using *pp_emg*. Do we need to keep them consistent? I guess *pp_scr* feels more natural if you don't mind...

teddychao:
I suggest the help text can be saved in a file and loaded when needed. This can help to make the text consistent throughout the software.

dominikbach:
The options of pspm_scr_pp are not accessible from the GUI. They were accessible when it was part of artefact rejection, so the old code can simply be copy-pasted.

I made two changes:
1. Help text in pspm_scr_pp streamlined.
2. Channel action default in pspm_emg_pp reverted - this should be discussed separately.

teddychao:
> The options of pspm_scr_pp are not accessible from the GUI. They were accessible when it was part of artefact rejection, so the old code can simply be copy-pasted.
> 
> I made two changes:
> 
> 1. Help text in pspm_scr_pp streamlined.
> 2. Channel action default in pspm_emg_pp reverted - this should be discussed separately.

Thanks Dom. I have made several modifications

* the GUI now has the parameters of pp_scr accessible, e.g. the min vale.
* Options for clipping detection have been available in the GUI.
* The main GUI has provided the access to *preprocessing scr*, both the old and new main GUI.

Many thanks again.

dominikbach (approved):

PR #197, opened 2021-02-23
Fix issue 196

### LINKED ISSUE #196: Improves dialogs
PsPM is currently use the function _menu_ for asking users to specify a choice. However, the _menu_ does not have a default return, which could bring risks to the program fallback, like endless loops. The function _questdlg_ (https://www.mathworks.com/help/matlab/ref/questdlg.html) recommended by mathwork can resolve this issue. It would be useful to check all related functions and update _menu_ to _questdlg_.

teddychao:
Fixes #196 .

- Make question dialog not shown in Jenkins (no window version)
- Change question dialogs from using function _menu_ to function _questdlg_

### DIFF
--- a/src/backroom/pspm_find_data_epochs.m
+++ b/src/backroom/pspm_find_data_epochs.m
@@ -105,8 +105,11 @@ if file_exist
     if options.overwrite 
         write_ok = true;
     elseif ~options.dont_ask_overwrite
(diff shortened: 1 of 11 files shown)


teddychao:
@dominikbach Apologies, the GUI description can be updated after understanding the full name of _scr_ in *pspm_scr_pp*.

dominikbach (approved):

Issue #185, opened 2021-02-06
Sub-session threshold and split error


teddychao:
Users had reported an error in _pspm_dcm_ when the imported data was split according to the sub-session threshold. Further investigating should be done for _pspm_dcm_ to see why this error is generated and how to make it able to process multiple sub-sessions.

PR #204, opened 2021-03-20
Fix Issue 185: Use master branch to test unexpected trials

teddychao:
Testing branch based on the master branch for fixing the issue #185  .


teddychao:
The issue is caused by the multiple sub-sessions.
Sometimes users wish to apply for a thresholding and build up multiple sub-sessions. In this occasion, a cell array containing multiple sub-sessions will be generated. However, such cell arrays cannot be used for the function cell2mat.
The current solution is to manually convert the cell array into a matrix for the following processing.
The current solution is able to resolve the issue, for both the main branch and the develop branch.

dominikbach:
@teddychao 
I noted the solution is in a different branch (208).
Is this one (204) a testing branch that can be closed/deleted without merging?
Or does it have to be merged?

dominikbach:
Following comments on 208 I assume it can be closed without merge.

PR #215, opened 2021-05-03
Fix Issue 214

### LINKED ISSUE #214: Improve the compatibility of trim for its use by split_sessions
### Feature Description
- The current *trim* uses a universal pair of start and end points to process a series of data. However, users may need to use different start and end points for the datasets, e.g. for split sessions. The *trim* should be updated for being available with vectors of start and end points.
- The current parameter settings for *trim* are very different from *split_sessions*. It could be simplified for better used as a generalised tool.

teddychao:
Fixes #214  .

- Update the input values of start and end points into numbers or vectors.

This makes it easier for *split_sessions* to call *trim*.


dominikbach (commented):
I'm wondering how we can simplify the code (in general, and here in particular).
There are three different ways to call this function; they were historically created (actually, by myself), but I think only one of them (one file at a time) is generally used.
Would it not be sufficient if this function could be called for one file at a time?
Would that simplify the code, shorten the input checks, and facilitate future maintenance?
Or is there a compelling reason to 'vectorize' the function?

teddychao:
> I'm wondering how we can simplify the code (in general, and here in particular).

Basically I am investigating how this function was used and could be optimised. It could be simplified as it is expected to be a basic and general function. However, I am pulling this request to specify the *from* and *to* values. I am confused about these values because it looks like they are always set to be fixed values, even if a cell (multiple pieces of data) is used. When we call *split_sessions*, the *from* and *to* values are not always fixed, so that is why I pull this request. 

> There are three different ways to call this function; they were historically created (actually, by myself), but I think only one of them (one file at a time) is generally used.

I agree with that. Currently it looks like only the _struct_ is used, instead of loading a datafile. However, this function looks like being expected to handle datafilee mainly as it spend quite many steps for loading and determine parameters (or info).

> Would it not be sufficient if this function could be called for one file at a time?

Yes, that is possible, and it is also what I am doing. However, I do not know if it is good as there are quite many *for* loops.

> Would that simplify the code, shorten the input checks, and facilitate future maintenance? Or is there a compelling reason to 'vectorize' the function?

I would suggest that too. It looks like only a struct is used (by all functions that use *trim*) currently. Can I create a simplified version of *trim* that handles only *struct*, which can be called by *split_sessions* and also other related functions?

dominikbach:
1. I wasn't aware that the struct input is actually needed, so this will have to be kept then.
2. When calling pspm_trim from the matlabbatch, it would be preferable if a string input (filename) was provided and let pspm_trim do the I/O handling using pspm_load_data (if that is not the case anyway). Using a filename is also useful when directly calling pspm_trim from the command line (as is for example done in the pspm_pipeline on gitlab).
3. The cell array input can probably be removed. Perhaps we can keep if for now, for legacy reasons? I'm always worried about touching a function that works.
4. pspm_split_sessions could then call pspm_trim one file at a time, and the "from" and "to" arguments would not need to be changed.

teddychao:
Trim is now compatible with Split Sessions. Optimisation could be made in the future, like handling struct or file.

dominikbach (commented):
After approving the fix of pspm_split_sessions, there is now a conflict in this branch. Could you kindly resolve?

dominikbach (approved):

PR #217, opened 2021-05-04
Fix Issue 216 (legacy)

### LINKED ISSUE #216: Test function generates ImportTestData at the upper level of folder
The code attempts to search for the folder *ImportTestData* at the upper level, if *ImportTestData* is missing in the current folder. However, all activities shall be put under the current home folder.
If it is set for optimising the Jenkins workplace, a patch can be added.

teddychao:
1. The name of WCHN should be Wellcome Centre for Human Neuroimaging.
2. the function *display* is discouraged, while *disp* or *fprintf* is encouraged.
3. If the folder *ImportTestData* cannot be found, the program shall ask the user to specify the path, instead of looking for that at the upper level of folder.

Fixes #216 .

### DIFF
--- a/src/pspm.m
+++ b/src/pspm.m
@@ -2,7 +2,7 @@ function varargout = pspm(varargin)
 % PsPM is the main GUI for PsPM
 %__________________________________________________________________________
 % PsPM 3.0
-% (C) 2008-2015 Dominik R Bach (Wellcome Trust Centre for Neuroimaging)
+% (C) 2008-2015 Dominik R Bach (Wellcome Centre for Human Neuroimaging)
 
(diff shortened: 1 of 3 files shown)


dominikbach (approved):
I don't have a comment; I haven't worked with the test system for a long time.

teddychao:
This branch is always reporting error in my computer, but results show no error in Jenkins. I do not quite understand why. I have reopen a pull request and that one runs well.

teddychao:
It looks like fprintf cannot handle some output well. 
The function *display* in *test* could be updated in the future, but may not be of high priority currently.

Issue #209, opened 2021-04-12
Failure to import BrainVision data


Maren186:
Hello,

I've been trying to import BrainVision data using the PsPM Batch Editor. 
I specified the Data File (.eeg) and tried both to import the SCR Channel alone and also different Channels (SCR, ECG, Startle) at the same time. 

I unfortunately always got the same error message: 
12-Apr-2021 10:31:54 - Failed  'Import'
Unrecognized function or variable 'filename'.
In file "C:\Users\...\PsPM_5.0.0\Import\fieldtrip\fileio\ft_read_header.m" (v395), function "ft_read_header" at line 114.
In file "C:\Users\...\PsPM_5.0.0\pspm_get_brainvis.m" (???), function "pspm_get_brainvis" at line 28.
In file "C:\Users\...\PsPM_5.0.0\pspm_import.m" (???), function "pspm_import" at line 201.
In file "C:\Users\...\PsPM_5.0.0\pspm_cfg\pspm_cfg_run_import.m" (???), function "pspm_cfg_run_import" at line 118.
The following modules did not run:
Failed: Import

I couldn't find the solution for my problem and unfortunately importing the file manually through the Command Window did not work either.
I would be very thankful if you can help me with my problem or give me some general advice about how to use PsPM to properly import BrainVision data files.

Thank you very much. 
Best regards, Maren 

### Technical Info
* PsPM version: Version 5.0.0 (21.08.2020)
* MATLAB version: Matlab 2021a

teddychao:
Dear @Maren186 

Thank you for the description of this issue. We have understood the issue mentioned in this comment.
Could we have some sample data so we can replicate this issue and find the reason?

Kind regards
Teddy

Maren186:
Dear Teddy,

Thank your for your quick response!
How can I send/give you example data? (My file is 50.000 KB big and too big to be attached..)

Kind regards
Maren

teddychao:
Dear @Maren186 

Thanks for your patience and the provided files.
Based on the given documents and the version of software, the reason that leads to the issue is that, the provided files for BrainVision data importing. The current version of PsPM would expect that, if Brainvision_eeg is imported, a file of .egg and a file of .vhdr should be provided together. I checked the provided data and I guess .vhdr was not provided in the target folder where .eeg file was placed.
[...]

PR #219, opened 2021-05-07
Issue 209

teddychao:
Updates for #209 .

Users tried to use PsPM to import eeg data with extensions *amrk* and *ahdr*, while PsPM is currently expecting the *vhdr* and *vmrk* files. Updates were made for reminding users to use *vmrk* and *vhdr* files. However, it could be discussed if *amrk* and *ahdr* are expected to be applicable in the next generation of PsPM.


dominikbach (commented):
Agree with the change in processing biosignal files.

A few notes:

- what do the pspm_report and pspm_app_beta functions do?
- the show_help and show_forum functions are included here; perhaps they belong into a different pull request (and that is why the other one did not work?)

dominikbach (approved):

PR #222, opened 2021-05-08
Fix Issue 220

### LINKED ISSUE #220: Data editor not showing missing epochs properly
### Summary
Some functions, including loading the missing epochs, are not working well in data editor.
Data editor is also being converted into App-Design-based GUI, while some features, including showing missing epochs, seem to be not working either.

### Technical Info
* PsPM version: 5.0
* MATLAB version: 2021a

teddychao:
Fixes #220 .

- Update the GUI of data editor.
- Update the function of some buttons.


teddychao:
I am a little confused about the current GUI.
It looks like the two buttons at the bottom right corner have the same functions to the system ones.
- The cancel button works equivalently to the close button of the window.
- The apply button works equivalently to the 'save output file' button.
May I know if this is functioning as expected? @dominikbach

dominikbach:
> I am a little confused about the current GUI.
> It looks like the two buttons at the bottom right corner have the same functions to the system ones.
> 
> * The cancel button works equivalently to the close button of the window.
Users may prefer a button to end the data editor rather than closing the window; but it could be renamed "quit"?
> * The apply button works equivalently to the 'save output file' button.
I've been wondering about this as well; I'm not sure this was done on purpose. 
> 
> May I know if this is functioning as expected? @dominikbach

teddychao:
The function seems to originally designed for showing only one epoch, because the epochs are read and then covered by new epochs.. Since users may need to see the specified epoch, it is currently being rewritten...

teddychao:
Should be working now...

dominikbach:
I'll comment after Juliana tested the functionality.

teddychao:
I have fixed this issue
* The missing epochs were not displayed when the data editor was run from the command window.

However, I am confused about the old code. It seems like the missing epochs sections and export results sections in the GUI were not expected to be displayed if the data editor was run from the command window. Although I understand the logic, as users have specified their path for missing epochs and exported results, I am not certain that this is what is expected for data editor.
The current code would always show the path for missing epochs and exported results, no matter it is run from the command window or manually operated.
May I know your ideas? @dominikbach @jusporrer 

Thanks.

teddychao:
I think the current list of epochs is a little confusing. In English, when we would like to express the period from A to B, we normally use the en dash as A--B instead of hyphen symbol A-B ([Reference](https://en.wikipedia.org/wiki/Dash#En_dash)). I also added the space to make the list look tidy.

teddychao:
I have raised a new issue regarding the order of the listed missing epochs, because I am not sure if that is emergent. If we could approve this pull request and merge it to the develop branch, it will be more convenient for Juliana's work I guess.

dominikbach:
> I have fixed this issue
> 
> * The missing epochs were not displayed when the data editor was run from the command window.
> 
> However, I am confused about the old code. It seems like the missing epochs sections and export results sections in the GUI were not expected to be displayed if the data editor was run from the command window. Although I understand the logic, as users have specified their path for missing epochs and exported results, I am not certain that this is what is expected for data editor.
> The current code would always show the path for missing epochs and exported results, no matter it is run from the command window or manually operated.
> May I know your ideas? @dominikbach @jusporrer
> 
> Thanks.

@jusporrer ?

dominikbach:
Talked to @jusporrer and we agreed that the path should always be displayed and editable, no matter how the data editor is called.

dominikbach (approved):

PR #230, opened 2021-05-21
Update UI

teddychao:
Users have reported the text in the new UI is too small for them. This pull request updates the UI for the release 5.1.0, specifically optimised for Windows 10.

- Update main screen and display


teddychao:
The UI is updated based on the following rules
- Titles are set to be 9 points with bold weight
- Main text or options are set to be in 8 points with normal weight
- Attribution or smaller text (like *status*) are set to be in 7 points with normal weight

All text are set to be in Verdana

dominikbach:
Somehow it appears really small on my Mac - smaller than with the version in the develop branch. It seems the small fonts ("Data Preparation", etc.) were bold before and are not (or are just smaller) now? Even before the fonts were quite small. I would suggest making it all a bit larger?

teddychao:
> Somehow it appears really small on my Mac - smaller than with the version in the develop branch. It seems the small fonts ("Data Preparation", etc.) were bold before and are not (or are just smaller) now? Even before the fonts were quite small. I would suggest making it all a bit larger?

I met the same problem on my computer. The design in the develop branch is optimised for Mac, while the text cannot be displayed completely on a Windows machine. It looks like the logical resolution for the GUI between Mac and Windows is different. I guess a possible solution is add manual specification for Mac, because windows users would be the majority. I am trying to add such specification atm. 

By the way, it looks like the app designer based GUI is consistent between Mac and Windows, presumably also in Linux.

teddychao:
It looks like a [factor](https://uk.mathworks.com/matlabcentral/answers/378936-windows-mac-font-size-differences-have-become-unworkable) should be added to fix the huge font in windows version. Can be tried tomorrow when I am in the BCH office.

teddychao:
I have set the GUI in a normalised size. According to the official document of MATLAB, this design could make the GUI look universally similar between different operating systems. 

I also set the window to be proportional, indicating that users can adjust the window size if they want it to be bigger, especially on Windows.

I will see if I can adjust it again in the next couple of days. I guess it would be useful to add this into the develop branch so that other group members can give their feedback to this version.

dominikbach (approved):

SerenXia (approved):
The UI looks all good from my side.
The original size called by "pspm" was a bit small (see attached figure), but the good thing is that I can not adjust the size myself (see attached figure).
pspm_gui_original_size
pspm_gui_resized

Issue #251, opened 2021-06-18
Refactoring of  pspm_split_sessions


dominikbach:
### Summary
pspm_split_sessions was refactored to use pspm_trim. However, it calls pspm_trim for every channel, which is unnecessary and error-prone. Also, it retains the bug "makerinfo" instead of "markerinfo".

### (Optional) Possible Cause and Solutions
Call pspm_trim once for each session, rather than for each data channel.

PR #252, opened 2021-06-18
fix bugs in split_sessions

### LINKED ISSUE #250: Overwrite Bug in pspm_split_sessions
### Summary

pspm_split_sessions does not use the argument options.overwrite.

### Steps to Reproduce
1. Set options.overwrite = 1
2. Run pspm_split_sessions
3. Run again, thus seeking to overwrite files

### Expected Results
Files overwritten without question dialogue 

### Actual Results
Question dialogue pops up

exist('options.overwrite','var') in line 85 is always false (because options.overwrite is not a variable)

Solution:
if ~exist('options','var') || isempty(options) || ~isstruct(options)
    options = struct();
elseif ~isfield(options,'overwrite')
    options.overwrite = 0;
elseif options.overwrite ~= 1
    options.overwrite = 0;
end

### Technical Info
* PsPM revision ID: 5.1.0
* MATLAB version: 2019b

dominikbach:
Fixes #250, #251, #254 

- fix bug with overwrite
- enable full use of pspm_trim, thus slashing down code volume in pspm_split_sessions
- this removes a remaining bug (typo: makerinfo instead of markerinfo)

### DIFF
--- a/src/pspm_split_sessions.m
+++ b/src/pspm_split_sessions.m
@@ -22,7 +22,7 @@ function [newdatafile, newepochfile] = pspm_split_sessions(datafile, markerchann
 %                           in terms of markers (vector of integer)
 % options.prefix            In seconds, how long data before start trim point
 %                           should also be included. First marker will be
-%                           at t = -options.prefix
+%                           at t = options.prefix
 %                           Default = 0
 % options.suffix            In seconds, how long data after the end trim
 %                           point should be included. Last marker will be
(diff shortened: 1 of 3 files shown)


teddychao (approved):

SerenXia (approved):
The current version works perfectly with data files that contained splitting issues using the previous split function.

Also tested with some data files without issues previously and everything went well with those data using the current function.

PR #267, opened 2021-06-27
Fix Issue 266

### LINKED ISSUE #266: Errors when running pspm_dcm with missing epochs
### Summary
Several types of errors occur when running pspm_dcm with missing epochs.

1. "Matrix dimensions must agree" in line 598 of pspm_dcm
2. "Unable to perform assignment ..." in line 613 of pspm_dcm

### Steps to Reproduce
1. @jusporrer provided several example data sets on a private repository

### Expected Results
No errors

### Actual Results
Errors

Wrong indexing for missing epochs or trials.

### Technical Info
* PsPM revision ID: v5.1.1 (latest release)
* MATLAB version: ?

teddychao:
Fixes #266 .

- The undefined variable srscr in line 167 might be presumably missingsr.


teddychao:
I manually updated the parameters for processing missing epochs. I do not know why the marker information was added to dummy data{2}, so I temporarily disabled it. The code is currently running ok. @jusporrer Could you please have a look and check if it functions as you expect, thanks 😉

dominikbach:
@teddychao could you test the issue on Juliana's data again? In general, when you detect bugs in my code it may save you time to assign the issue to me (or get in touch). You had disabled the marker channel to be used for splitting missing epochs, which made the errors disappear but also removed the functionality.

teddychao:
@dominikbach I am a little confused about the current logic.
1. The option of processing epochs in _trim_
At the beginning, I saw the option for trim's epochs processing was set to 'custom'. I am not sure if that means I need to select an option between 'marker' and 'file' at this step. Based on my understanding and previous code, I think it should be 'marker', if no new option is defined in _trim_, because the data to process in this step contains both the marker channel and the epochs, and the former requires the option 'marker'.
2. Based on step 1, where both the marker channel and the epochs should be processed with the option 'marker', _trim_ will use the data for setting the start point. In line 253 in _trim_, because 'marker' is selected, the start point is defined as sta_p = events(startmarker). However, this is not correct for missing epochs that are dominated by 0, because the events will be generally 0 and the start point will be probably 0 too. For example, like Juliana's situation, the event at that startmarker is 0, so the start point is consequently 0. With the previously defined shift, the start point will be -5. It seems a little wired in my opinion.
3. I think consequently we may need to build a new option in _trim_ for processing epochs. For a struct of epochs, it contains both the missing epochs data and the marker channel data. The latter would have no problem, because the logic is consistent with the previous step of processing the main data. However, the former would be tricky, because the logic is different for missing epochs. This is why I proposed to modify _trim_ by adding a new option for epochs.

teddychao:
The error happened to flags has been resolved by updating cell array handling. However, the code is throwing warnings because the end point has exceed the length of data. I am not sure this is due to the parameter settings. If the trim parameters are correct, I will go back to check if the split_sessions. I am still a little confused as the warning happens to only epochs but not appears for the raw data.

dominikbach:
Thanks @teddychao for spotting the bugs in pspm_split_sessions. I made the following additional changes:
1. Removed repeated checking of the input argument options.missing - this needs to be checked only once.
2. Removed warnings if options.missing does not exist.
3. Write out missing files even if they are empty.
4. Removed the option to use a cell array of data file names - this is unnecessary, and also it would require that we allow options.missing to contain a cell array of equal size. 
5. Corrected the indexing of the missing epochs dummy channel in several places.

I have checked it on dummy data sets but could you please also check on Juliana's data.
Also, could you please check that matlabbatch allows only one data filename rather than a cell array.

I will check the dcm implementation once we have finished this.

dominikbach:
checked Juliana's data set 6 - ok.

teddychao:
@dominikbach Thanks a lot for the suggestions on the code, I agree that some of them seems a little redundant. I have checked Juliana's dataset, here are the results
- 6, 18, 19: throw warnings that the trimimg end points is out of the length and the defined epoch is empty.
- 11: throw warnings that the defined epoch is empty and the trial 14 requires NaN filling.

dominikbach:
Revert to previous state: no cell array input allowed in pspm_split_sessions.

All warnings in Juliana's data are valid.

However, in data set 11 I discovered another inconsistency in pspm_dcm (last trial filtering is done twice, with different results, only one is used), which I will try to fix now.

dominikbach (approved):

dominikbach:
Thanks for the review, @teddychao. The refactoring made the code shorter and removed bugs in some edge cases. Ready for merging now. Summary of changes:

(1) Identification of missing epochs updated: session offset removed.
(2) Last trial filtering refactored. Was done in two places, and one was not used. The one that was used only detected last trials in user-defined sessions, not in subsessions created by missing epochs. This is now unified with the detection of trials that lie within missing epochs (using the variable trials{sn}). 
(3) Comments and help text updated.

dominikbach:
Thanks for the review, @teddychao. The refactoring made the code shorter and removed bugs in some edge cases. Ready for merging now. Summary of changes:

(1) Identification of missing epochs updated: session offset removed.
(2) Last trial filtering refactored. Was done in two places, and one was not used. The one that was used only detected last trials in user-defined sessions, not in subsessions created by missing epochs. This is now unified with the detection of trials that lie within missing epochs (using the variable trials{sn}). 
(3) Comments and help text updated.

teddychao:
@dominikbach I really agree with the updates in this version. Many thanks.
Just to mention, I tested juliana's dataset just now. Subjects 11, 18 and 19 reported the use of `Filling the rest with NaNs`. I am not sure if that is expected...

dominikbach:
Yes, this is expected when the last trial is closer to the session end than the minimum iti in all other trials of this session; this is the case here.

dominikbach:
I believe the $rev$ is a legacy from times when we used subversion for version control. I don't think it is updated in git. Hence, the dcm.revision does not contain meaningful information, and we can remove it?

teddychao:
@dominikbach Yes, and the only concern is I do not know if `rev` means `code review` or `code revision`. If not, I think we can remove these `rev`.. I was confused as I found two parallel version control names: the rev ID and the current version number (e.g. 5.0.0), and the latter is what we show to users when initialising PsPM. If useful, I can pull a request to generalise the version control information throughout the software..

dominikbach:
This referred to code revision. $Rev:$ invokes a keyword substitution in subversion, and writes the number of the current revision into the file. We don't have this feature any more, so if we want to keep track of the current version of bugtracking purposes then this would need to be solved differently. I'm not sure though we need this - the users can usually tell us which version they used.

PR #271, opened 2021-07-01
Add a new feature in return values of pspm_load_data (channel index)

dominikbach:
Minor change to pspm_load_data: return index of the channels that are returned. Useful for functions that only accept channel numbers rather than channel types (e.g. pspm_extract_segments).

### DIFF
--- a/src/pspm_load_data.m
+++ b/src/pspm_load_data.m
@@ -42,12 +42,12 @@ function [sts, infos, data, filestruct] = pspm_load_data(fn, chan)
     %                       .numofeventchan
     %                       .posofmarker' (first marker channel, or 0 if no 
     %                          marker channel exists)
+    %                       .posofchannels (number of the channels that
+    %                       were returned)
     %__________________________________________________________________________
-    % PsPM 3.0
-    % (C) 2008-2015 Dominik R Bach (Wellcome Centre for Human Neuroimaging)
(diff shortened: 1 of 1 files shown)


src/pspm_load_data.m
```diff
@@ -253,15 +253,20 @@
...
     elseif isnumeric(chan)
         if chan == 0, chan = 1:numel(data); end
         if any(chan > numel(data))
             warning('ID:invalid_input', 'Input channel number(s) are greater than the number of channels in the data'); return;
         end
         data = data(chan);
+        filestruct.posofchannels = chan;
     elseif isstruct(chan) && ~isempty(fn) && (~exist(fn, 'file') || chan.options.overwrite == 1)
         save(fn, 'infos', 'data');
+        filestruct.posofchannels = 1:numel(data);
```
teddychao, line 267: Could I know if this sentence means all channels will be used?

dominikbach: In this case, pspm_load_data checks & saves data that have been passed over as a struct. It then returns everything that was passed as an input, and so for consistency I wanted to populate the fields anyway.

teddychao (approved):

teddychao:
Jenkins is not working for this pull request.. adding a new commit will recall Jenkins to run it, for example, add a single space...

PR #274, opened 2021-07-02
Fix-272

### LINKED ISSUE #272: Split sessions file loading issue
### Summary

Split sessions should accept data file and missing epoch file as char variables from either GUI or scripts. However, users reported currently the missing epoch file was not accepted sometimes. It is useful to check if the information of data file and missing epoch files are delivered correctly to the function.

dominikbach:
Fixes #272  .

- allow splitting of empty missing epochs in pspm_split_sessions for compatibility with pspm_scr_pp

### DIFF
--- a/src/pspm_split_sessions.m
+++ b/src/pspm_split_sessions.m
@@ -1,8 +1,7 @@
 function [newdatafile, newepochfile] = pspm_split_sessions(datafile, markerchannel, options)
 % pspm_split_sessions splits experimental sessions/blocks, based on
 % regularly incoming markers, for example volume or slice markers from an
-% MRI scanner (equivalent to the 'scanner' option during import in previous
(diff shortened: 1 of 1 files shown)


teddychao (approved):

src/pspm_split_sessions.m
```diff
@@ -166,8 +166,11 @@
         missing = round(missing*missingsr); % convert epochs in sec to datapoints
     end
     indx = zeros(1,round(missingsr * ininfos.duration)); % indx should be a one-dimensional array?
-    indx(missing(:, 1)) = 1;
-    indx(missing(:, 2)+1) = -1;
+    % allow splitting empty missing epochs
```
teddychao, line 169: I agree the old code always take effects on the missing epochs, which is not expected. It should be applied to missing epochs when users define missing epoch files.

PR #290, opened 2021-07-30
Fix issue 275

### LINKED ISSUE #275: Integrate pspm_find_valid_fixations with pspm_pupil_pp
### Feature Description

pspm_find_valid_fixations should allow:
- taking a preprocessed pupil channel as input
- allow a r/l-comined preprocessed pupil channel
- in this case, integrate r/l gaze channels with nanmean

SerenXia:
### Eye data pp

Fixes #275  .

see comments here: https://github.com/bachlab/PsPM/issues/275#issuecomment-889949096


teddychao:
Changes in this pull request

1. `pspm_find_valid_fixations` used to accept `gaze_x`, `gaze_y`, `pupil`, `pupil_missing`, and `left` / `right` is determined according to the variable `eye`. Now `pspm_find_valid_fixations` should be able to accept the combined `left-right`, which is also determined in `eye`.

src/pspm_init.m
```diff
@@ -411,6 +411,58 @@
     'import', @none, ...
     'data', 'wave');
 
+% documented
+defaults.chantypes(37) = ...
+    struct('type', 'gaze_x_lr_pp', ...
```
teddychao, line 416: @SerenXia Because some code seems to use one character to refer one side of the eyes, I do not recommend to use `lr` to denote "the combination of left and right eyes". I temporarily change it to `b` which denotes "bilateral". It would be easier to maintain some code, if you do not mind. We can change it back if it is preferred to use `lr` anyway. Thanks.

src/pspm_init.m
```diff
@@ -192,285 +203,285 @@
...
+  struct('type', 'custom', ...
+  'description', 'Custom', ...
+  'import', @pspm_get_custom, ...
+  'data', 'wave');
 defaults.chantypes(34) = ...
-    struct('type', 'pupil_r_pp', ...
-    'description', 'Pupil right preprocessed', ...
-    'import', @none, ...
-    'data', 'wave');
+  struct('type', 'pupil_r_pp', ...
```
teddychao, line 409: @SerenXia Sorry, I do not recommend to write it like `pupil_r_pp`, instead, write it as `pupil_pp_r`, because some code assume the "side marker" is written at the end, for example in line 331--332 in `pspm_find_valid_fixations`.
> `channels(str_chans) = regexprep(channels(str_chans), '(pupil_pp|gaze_x|gaze_y|pupil_missing)', ['$0_' eye]);`

SerenXia: > @SerenXia Sorry, I do not recommend to write it like `pupil_r_pp`, instead, write it as `pupil_pp_r`, because some code assume the "side marker" is written at the end, for example in line 331--332 in `pspm_find_valid_fixations`.
> 
> > `channels(str_chans) = regexprep(channels(str_chans), '(pupil_pp|gaze_x|gaze_y|pupil_missing)', ['$0_' eye]);`

Hi Teddy, this naming format was not determined by me. I don't know who determined it, but if you want to change them (as in many related functions they are named in this way), you may need to discuss with Dominik.

Best regards,
Yanfang

src/pspm_init.m
SerenXia: > @SerenXia Because some code seems to use one character to refer one side of the eyes, I do not recommend to use `lr` to denote "the combination of left and right eyes". I temporarily change it to `b` which denotes "bilateral". It would be easier to maintain some code (e.g. line 173--188 in `pspm_find_valid_fixations_test`), if you do not mind. We can change it back if it is preferred to use `lr` anyway. Thanks.

Hi Teddy, again, this naming format was not determined by me. If you would like to change, you may need to discuss with Dominik.

Best regards,
Yanfang

src/pspm_init.m
teddychao: > > @SerenXia Sorry, I do not recommend to write it like `pupil_r_pp`, instead, write it as `pupil_pp_r`, because some code assume the "side marker" is written at the end, for example in line 331--332 in `pspm_find_valid_fixations`.
> > > `channels(str_chans) = regexprep(channels(str_chans), '(pupil_pp|gaze_x|gaze_y|pupil_missing)', ['$0_' eye]);`
> 
> Hi Teddy, this naming format was not determined by me. I don't know who determined it, but if you want to change them (as in many related functions they are named in this way), you may need to discuss with Dominik.
> 
> Best regards,
> Yanfang

Exactly what I am going to do!
Thanks Yanfang.

Best
Teddy

teddychao:
Hi @SerenXia Just have a question about `best eye`. I remember you mentioned
> use the left eye, the right eye, the best eye, or combine two eyes

Can I confirm with you that `the best eye` should be a basic situation, like `left eye` for example? The code is currently setting `left`, `right`, and `all` as the three basic situations, but `best eye` is a dependent situation and used only for some specific functions that include this feature. May I know your expectation please? Thanks

Best wishes
Teddy

SerenXia:
> Hi @SerenXia Just have a question about `best eye`. I remember you mentioned
> 
> > use the left eye, the right eye, the best eye, or combine two eyes
> 
> Can I confirm with you that `the best eye` should be a basic situation, like `left eye` for example? The code is currently setting `left`, `right`, and `all` as the three basic situations, but `best eye` is a dependent situation and used only for some specific functions that include this feature. May I know your expectation please? Thanks
> 
> Best wishes
> Teddy

as far as I know, after importing eye data, pspm will access the quality of data from both eyes, and store the "best eye" in info.source.besteye (the eye containing less missing data is regarded as the best eye)

dominikbach:
@teddychao A few questions and points to make sure I understand this pull request.

1. Why do we need preprocessed gaze channels? There is no function that creates or uses them, so can we remove them for now, or do we want to keep them for the future?
2. Combined channels: new identifier "b" not yet implemented in pspm_init?
3. pspm_load_data: precedence order for sps should only account for combined - best since there are no pre-processing functions available
4. Is there a function to combine sps or gaze for left-right?
5. How does find_valid_fixations work - does it combine gaze channels on-the-fly when using combined pupil data?

SerenXia:
> @teddychao A few questions and points to make sure I understand this pull request.
> 
> 1. Why do we need preprocessed gaze channels? There is no function that creates or uses them, so can we remove them for now, or do we want to keep them for the future?
> 2. Combined channels: new identifier "b" not yet implemented in pspm_init?
> 3. pspm_load_data: precedence order for sps should only account for combined - best since there are no pre-processing functions available
> 4. Is there a function to combine sps or gaze for left-right?
> 5. How does find_valid_fixations work - does it combine gaze channels on-the-fly when using combined pupil data?

Dear Dominik,

Regarding point 1, 3, and 4, we would like to keep preprocessed gaze channels for the future functions to combine sps or gaze for left-right.

Regards,
Yanfang

dominikbach (approved):

PR #292, opened 2021-08-12
Fix issue 291

### LINKED ISSUE #291: Bug in pspm_display (heart beat channel)
### Summary
When plotting heart beat channel with no wave channels, the inter beat interval is plotted against the heart beat time stamps.

Values on the x-axis (time stamps) have larger range than the actual time stamps in the file - see screenshot below (duration 3300 s, plotted points go beyond 4000 s)

### Steps to Reproduce
Open file this file with pspm_display:
https://liveuclac.sharepoint.com/sites/BachlabUCL/Shared%20Documents/PsPM%20bug%20data/pspm_PCF3_037.mat

### Expected Results
X-axis values match timestamps in heart beat channel.

### Actual Results
X-axis values go beyond time stamps in heart beat channel.

### Screenshots if Graphical User Interface is Used

image

### Technical Info
* PsPM revision ID: 5.1.1
* MATLAB version: 2019b

teddychao:
Fixes #291 .

- Update `pspm_display.m`
  - `pspm_display` now sets limits for the x axis of the plot when the relative information is available from the datafile.


teddychao:
I do not know what Apple does in macOS 11.5.1, but it seems that light/italic/bold variations of fonts are no longer accessible in MATLAB (e.g. Helvetica Light is not displayed). I temporarily change *Helvetica Light* to *Gill Sans* for emphasising. Hopefully Apple or Mathworks can fix that in the future.

teddychao:
image
I think this is theoretically what is expected. I also changed the x axis label to *Time [s]* because the original *Time in second [s]* looks a little wordy.

dominikbach:
@teddychao 

I agree with the limits. 

But I'm wondering why data points were displayed at x-values that did not exist. Is this fixed, too?

I agree that the tick labels are quite small and could be increased.

teddychao:
> But I'm wondering why data points were displayed at x-values that did not exist. Is this fixed, too?

Thanks for pointing out this issue. I checked the imported data just now, and I found the variable `wave` became around 13.5. I am not sure if such information should be removed or not. There is only one `0` at the end. Here are the results

image

image

teddychao:
@dominikbach I open a new issue about UI parameters (#293 ). It would be easier for any function to get those UI parameters for plotting. The current `pspm_ui` works only when initialising a window, thus its parameters cannot be used by other functions.. May I know if you are happy with that? Thanks

dominikbach:
@teddychao What I meant was: in the screenshot in issue #291, you see data points around 4000 s - they should not be there. This is independent of the axis limits.

teddychao:
> @teddychao What I meant was: in the screenshot in issue #291, you see data points around 4000 s - they should not be there. This is independent of the axis limits.

Yes the screenshot in #291 showed the data points around 4000s, and now it is at around 3300s [here](https://github.com/bachlab/PsPM/pull/292#issuecomment-897831981), may I know if this is better?

dominikbach:
But what you are plotting here are the SCR waveform data, not the HB data?

teddychao:
image
Yes, I only showed the the SCR waveform data, not the HB data. I checked the previous screenshot and plot it again showing only hb data, which seems to be different from the previous screenshot?.. However I agree I do need to check different types of data to be plotted.

teddychao:
@dominikbach My apologies, I did not understand the issue correctly. I have been aware aware of the issue and rewritten some code to make `display` function correctly, especially when only events are selected without any wave channels.

teddychao:
## selecting only marker
image
## selecting only hb
image
## selecting scr with marker
image

dominikbach (approved):

PR #299, opened 2021-08-27
Fix issue 298 (legacy, not used)

### LINKED ISSUE #298: Incorrect parameter setting in GLM for scanpath-speed
### Summary

The default setting of `pspm_bf_spsrf_box/gamma` was not correctly set in `pspm_init`. 
In pspm_init, GLM modality settings: previously in subfield 'cbf', the value of 'args' was set to be 1. This is not correct.
If args is set to be 1, the default SOA will be 1, rather than 3.5 s (as set in pspm_bf_spsrf_box/gamma). 'args' should be empty, so that the basis function will take 3.5 s as default SOA. Changing SOA with specific value during model inversion works perfectly.

---
On behalf of @SerenXia

teddychao:
The default value is updated from 1 to 3.5.

Fixes #298 .

- Update the default value of GLM for Scanpath-speed, from 1 to 3.5.


src/pspm_init.m
```diff
@@ -989,22 +989,22 @@
...
     'default', 1);
 
 defaults.glm(10) = ... % GLM for Scanpath-speed
-    struct('modality', 'sps', ...                               % modality name
-    'modelspec', 'sps', ...                                  % modality name
-    'cbf', struct('fhandle', @pspm_bf_spsrf_box, 'args', 1), ...  % default basis function/set
-    'filter', struct('lpfreq', NaN, 'lporder', NaN,  ...        % default filter settings
+    struct('modality', 'sps', ...
+    'modelspec', 'sps', ...
+    'cbf', struct('fhandle', @pspm_bf_spsrf_box, 'args', 1), ...
```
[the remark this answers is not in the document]
teddychao: Hi, thank you for the explaining. I think I have got more understanding of the issue in this model. I think both `pspm_init` and `pspm_bf_spsrf_box` (or together with other potential similar functions) shall be updated. However, I am not sure if this feature is included in the test function correctly. May I know if you mind sending me one or two lines of code showing how this model is used by `pspm_bf_spsrf_box`? This would make it easier for me to update the test function as well. Many thanks!

teddychao:
@SerenXia I think I have made some progress which is closer to what you suggest. May I ask you to check `pspm_bf_spsrf_box` and let me know if you are happy with it? If it works well, I will do the following for other related functions, including the test functions. Many thanks.

src/pspm_bf_spsrf_box.m
```diff
@@ -1,60 +1,64 @@
...
 elseif nargin == 1
-    n_el = numel(varargin{1});
-    td = varargin{1}(1);
-    if n_el > 1, soa = varargin{1}(2); else , soa=3.5; end;
+  n_el = numel(varargin{1});
+  td = varargin{1}(1);
+  if n_el > 1
+    soa = varargin{1}(2); 
+  else
+    soa = settings.glm(10).filter.soa;
```
SerenXia, line 32: no need to change like this. the values are the input arguments (varargin) when calling this function. Just keep it as it was, so the default SOA will be 3.5. 

when we want to do model inversion (i.e. run glm), we first set model information.
datafile = .... (filename)
timing = ...(timing for glm);
soa = ...(experimental soa, if not set, the function should take it as default, i.e. 3.5 s here);
mod = 'sps';
outputfile = ... (the output filename after glm);

% option 1: now set the model for glm without specific SOA, but just take the default RF and SOA
model.timing = timing;
model.modelspec = mod;
model.modelfile = outputfile;
model.datafile = datafile;
model.timeunits = 'markers'; %(i.e. timing according to trial markers)
options = struct('overwrite', 1);
mdl = pspm_glm(model, options);

the codes then go to pspm_glm, there in the function, line 348 % check & get basis function
it will check whether bf is pre-set or not. Here you can see that it takes settings.glm(modno).cbf;
and then later in line 373, it uses model.bf.args(:) as input argument for this bf function (i.e. pspm_bf_spsrf_box)
so please do not change anything in this bf function, only change the pspm_init is enough, as I suggested.

% option 2: now set the model for glm with specific SOA,
everything of model is the same as previous. But add the following:
model.bf.fhandle = @pspm_bf_spsrf_box;
model.bf.args = inarg.soa;

mdl = pspm_glm(model, options);

so now when the code runs into pspm_glm, it will take the soa I specifically set.

Conclusion:
if you want to change in your way, you need to change a bunch of codes and scripts. Could you please just simply change cbf settings of sps modality in pspm_init ?

dominikbach (changes requested):
I don't understand why so many changes are being made just to update the default parameter for the SOA?

Could you not just implement what @SerenXia suggested?

src/pspm_bf_spsrf_box.m
teddychao: Hi Yanfang, thank you for the very explicit explanation. Could I ask you to check the newly implemented solution [here](https://github.com/bachlab/PsPM/pull/307#issue-1000513029) and let me know if you are happy with it? Thanks.

PR #307, opened 2021-09-20
Fix issue 298

### LINKED ISSUE #298: Incorrect parameter setting in GLM for scanpath-speed

teddychao:
Fixes #298 .

- Update the default value for SOA in `pspm_bf_spsrf_box`.

Now `pspm_init` is updated and has the default value as 3.5 for SOA.

### DIFF
--- a/src/pspm_init.m
+++ b/src/pspm_init.m
@@ -1000,7 +1000,7 @@ defaults.glm(9) = ... % GLM for SEBR (fear-conditioning)
 defaults.glm(10) = ... % GLM for Scanpath-speed
     struct('modality', 'sps', ...                               % modality name
     'modelspec', 'sps', ...                                  % modality name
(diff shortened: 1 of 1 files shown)


SerenXia (approved):
Sure, if you would like this better.

dominikbach (approved):

PR #316, opened 2021-11-02
Testing GitHub actions with pushing data from another repository

teddychao:
Add ImportTestData from a private repo.


teddychao:
The current GitHub Actions workflow can read code from the current modification branch and data from the data repository. There are currently two issues:
* A private repository of test data is still not working. I am contacting the support of `actions/checkout` for solving this issue. 
* The result of GitHub Actions checking has not been returned back to the front end. I am checking how to do it and link it to GitHub's merging request.
This test workflow can be use for checking our modification branches before merging it into the develop branch, which is the same to Jenkins.

teddychao:
Because GitHub Actions restrict the free time for running test in each repo, I do some test [here](https://github.com/teddychao/PsPM/pull/1).

teddychao:
Good news! I have found a solution for running `pspm_test` on GitHub Actions. It runs perfectly and can replace Jenkins without missing features. I will update this pull request and finish all the GitHub Actions issues.

teddychao:
I would suggest to make the test function dynamic by letting it to test only the functions we are going to test. This is because GitHub Actions restrict the free time for running tests in each month. I will update it by adding the corresponding method.

teddychao:
@dominikbach This branch is what I used to test GitHub Actions. We have made up a solution which works for other pull requests, thus I will close this legacy pull request, sorry.

PR #319, opened 2021-11-13
Fixes issue 318

### LINKED ISSUE #318: Exceeded marker in importing viewpoint data
### Summary
During importing some viewpoint data, there is a bug reporting the marker has exceeded the limitation. This could be potentially due to the different line break format between Windows (CRLF) and Unix (Mac and Linux, LF). It is safer to correct the importing format by considering the situation when there are more information to be read.
image

### Technical Info
* PsPM version: 5.1.2
* MATLAB version: 2021b

teddychao:
Fixes #318.

- Update message detection by restricting the marker in `import_viewpoint_test`.


teddychao:
We have updated to resolve this issue in this pull request: https://github.com/bachlab/PsPM/blob/ef6f9b741b4ee36af5840c067eebcf8045f727d1/test/import_viewpoint_test.m

PR #327, opened 2021-11-16
Fix issue 326

### LINKED ISSUE #326: New function for converting time in second to data points
Time points were stored as seconds in some data (e.g. missing epochs), but they could be used as data points for processing or display. Converting these variables between seconds and data points needs generalisation throughout the software. It needs to notice that sometimes the converted value may exceed the length of data, which should be avoided in the upcoming conversion function.

teddychao:
Fixes #326.

- Add a new function for converting time-related variables


dominikbach (changes requested):
A couple of suggestions.
1. Two of the "options" fields are mandatory. I would suggest passing all mandatory arguments as separate arguments.
2. "time_conversion" sounds a bit unspecific to me, and it seems a bit cumbersome for typing. Would it not be easier to call it pspm_time2index in the firs place?
3. The reverse, index2time, could be a second function but since it just requires one line of code it might be easier to just write that code where it is needed?

teddychao:
I split the old `time_conversion` function into two separate functions `index2time` and `time2index`. `index2time` is now featuring an option of setting precision. It can be removed if it is not that essential as `time2index`.

dominikbach (changes requested):
Thanks for the changes. The code looks good to me.

I'd suggest clarifying the help text: both functions mention "time" and "index" as input argument, but only one of them is used - the other could be mentioned as output. 

Stylistically, I would also replace the "varargin" in the help text with "optional argument"

teddychao:
Thanks! I have made some modifications for the help text. Just can I confirm with you:
> both functions mention "time" and "index" as input argument, but only one of them is used - the other could be mentioned as output.
Can I know if you suggest to change it to this format: sts = pspm_time2index(time, index, sr, varargin)?
> Stylistically, I would also replace the "varargin" in the help text with "optional argument"
Yes, I have changed it into optional in help text. Just to explain that varargin is an optional argument which does not have to be provided. However, "options", previously used in other functions as a struct, must be provided. In that case, "options" can be an empty struct meaning no optional arguments were required. In this function I initially wrote "options" as a struct, but that may be a little wordy. Can I know if I will still use "varargin" instead of a struct but just mention they are options in help text?
If you are happy with this version of the code, I will update the help text in manual.
---
I have also provided two templates for writing functions. Can I know if you are happy with them? We can follow this style to write functions and generalise old functions if you don't mind.
Thank you.

dominikbach (approved):

PR #333, opened 2021-11-29
Fix issue 323

### LINKED ISSUE #323: pspm_split_sessions crashes when a missing file contains an epoch starting at 0 s
### Summary

When a missing epoch starts with "0", pspm_split_sessions crashes on line 171.

### Solution

Catch 0 values after line 168 and convert to 1.

teddychao:
Fixes #323 .

- `pspm_time2index`
  - It was updated to be able to accept data length as only data points (integers).
- `pspm_index2time`
  - It is removed since it is only one line of code.
- `pspm_split_sessions`
  - `pspm_split_sessions` was updated to use `pspm_time2index` for converting missing epochs.
  - `pspm_split_sessions_test` was reindented using MATLAB's own indent style and reformatted to utf-8, but not changed.
- `pspm_glm`
  - `pspm_glm` was reindented using MATLAB's own indent style and reformatted to utf-8.
  - `pspm_glm` now uses `pspm_time2index` under its new input rules.


src/pspm_split_sessions.m
```diff
@@ -154,17 +154,12 @@
...
         warning('ID:invalid_input', 'Could not load missing epochs.');
     end
     missingsr = 10000; % dummy sample rate, should be higher than data sampling rates (but no need to make it dynamic)
-    if any(missing > ininfos.duration)
-        warning('ID:invalid_input', 'Some missing epochs are outside data file.');
-        return
-    else
-        missing = round(missing*missingsr); % convert epochs in sec to datapoints
-    end
+    missing = pspm_time2index(missing_time, missingsr, ininfos.duration*missingsr); % convert epochs in sec to datapoints
```
dominikbach, line 162: could the last argument lead to problems as it is not an integer? result might also not be the same as in line 163 where ´´´round´´´is used?

teddychao: Obviously I made a stupid mistake... let me think how to make `pspm_time2index` easier to use

src/pspm_glm.m
```diff
@@ -1,931 +1,937 @@
...
+  % concatenate data
+  Y=[Y; NaN(newsr * model.bf.shiftbf, 1); newy(:)];
+
+  % get duration of single sessions
+  tmp(1).snduration(iSn) = numel(newy) + newsr * model.bf.shiftbf;
+
+  % process missing values
+  newmissing = zeros(size(newy(:)));
+  if ~isempty(missing{iSn})
+    missingtimes = pspm_time2index(missing{iSn},newsr,length(newmissing),'i');
```
teddychao, line 540: This is the only place that has been updated in `glm`.

src/pspm_split_sessions.m
dominikbach: I think the problem is that the duration is converted from a floating point number to integer in two different places: once in pspm_time2index, and once in line 163 with "round". The result is not very resilient: imagine somebody changes pspm_time2indx but forgets to update the "round" in line 163.

I think it would be much better if pspm_time2index would accept only an integer index duration, and the same integer would be used in line 163.

teddychao: I understand the issue now... obviously it is not necessary to calculate the duration twice. I have corrected this issue in the new pull request.

dominikbach (approved):

PR #334, opened 2021-11-29
Fix issue 324 - pspm_time2index

### LINKED ISSUE #324: pspm_dcm cannot handle missing epochs that start at 0 (or very close to zer0).
### Summary
Same as #323 but for pspm_dcm. Importantly, in both issues, zeros could be introduced by the rounding operation when the starting value of the missing epoch is very small (not necessarily zero).

### Solution
Catch zero values around line 370 and set to 1.

teddychao:
Fixes #324.

Changes proposed in this pull request: `pspm_dcm`
- `pspm_dcm` was reindented with MATLAB's own indent style and converted to utf-8 and LF.
- Update `pspm_dcm` to use `pspm_time2index`
  - pspm_time2index
    - Line 375
    - Line 443
    - Line 458
    - Line 756


teddychao:
@dominikbach Sorry I do not have some test data here... Theoretically the issue should be fixed since `pspm_time2index` will use a 1 instead of 0 for the specified situation.

teddychao:
@dominikbach Sorry Dom I just have a question about the converting function. We are currently using `round` based on our previous discussions. However there seem to be some other function used at the same time, like `ceil` in line 684 of pspm_dcm, `floor` in line 726 of `pspm_dcm`. Can I know if we would like to add an option of converting method in `pspm_time2index`? Or alternatively we want to use `round` in all these cases?

dominikbach:
> @dominikbach Sorry Dom I just have a question about the converting function. We are currently using `round` based on our previous discussions. However there seem to be some other function used at the same time, like `ceil` in line 684 of `pspm_dcm`, `floor` in line 726 of `pspm_dcm`. Can I know if we would like to add an option of converting method in `pspm_time2index`? Or alternatively we want to use `round` in all these cases?

I think in these cases the goal is different - you want to extract data with a window size based on the smallest recorded ITI (rounded down with floor) but make sure that any missing data in the last trial are properly filled with NaN - I would not touch this.

dominikbach (approved):

teddychao:
> > @dominikbach Sorry Dom I just have a question about the converting function. We are currently using `round` based on our previous discussions. However there seem to be some other function used at the same time, like `ceil` in line 684 of `pspm_dcm`, `floor` in line 726 of `pspm_dcm`. Can I know if we would like to add an option of converting method in `pspm_time2index`? Or alternatively we want to use `round` in all these cases?
> 
> I think in these cases the goal is different - you want to extract data with a window size based on the smallest recorded ITI (rounded down with floor) but make sure that any missing data in the last trial are properly filled with NaN - I would not touch this.

Thanks Dom. I checked the code and can confirm that only missing epochs are using `pspm_time2index` to convert variables.

PR #337, opened 2021-12-13
Fix issue 317 - pspm_interpolate

### LINKED ISSUE #317: Forced extrapolation in pspm_interpolate returns no results
### Summary
When pspm_interpolate forces extrapolation, no result is returned.

This is due to a wrong "return" in line 250 of pspm_interpolate.

teddychao:
Fixes #317 .

- Test allowing function to run when forcing interpolating

### DIFF
--- a/src/pspm_interpolate.m
+++ b/src/pspm_interpolate.m
@@ -1,10 +1,10 @@
 function [sts, outdata] = pspm_interpolate(indata, options)
 % pspm_interpolate
-% 
-% FORMAT: 
+%
+% FORMAT:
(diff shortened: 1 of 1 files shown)


teddychao:
@dominikbach Sorry Dom, I am a little confused about the issue. In the line 250, the code shows that function is returned as "Cannot extrapolate with method previous and overlap at the beginning." It looks like the reason why no results were returned. Currently I removed the two returns but I am not sure how you would expect it to run. Would you suggest that it is valid if "extrapolate with method previous and overlap at the beginning"? Thanks.

dominikbach:
as just discovered, had already been fixed with #331

PR #342, opened 2022-01-07
Fix issue 213 - design matrix

### LINKED ISSUE #213: plot of design matrix of pupil size glm with luminance as nuisance file
### Summary
1. PsPM GUI first-level model review should use glm.XM when plotting design matrix, but not glm.X
2. design matrix plot did not show the correct design matrix (see attachments), but the design orthogonality plot seemed correct
CFG_day1_ps_nuisance_design_matrix_0001
CFG_day1_ps_nuisance_design_orthogonality_0001
CFG_day2_ps_nuisance_design_matrix_0001
CFG_day2_ps_nuisance_design_orthogonality_0001
[mdl_ps_fc_tpspm_exampledata.zip](https://github.com/bachlab/PsPM/files/6414397/mdl_ps_fc_tpspm_exampledata.zip)

### (Optional) Screenshots if Graphical User Interface is Used
please see attachments

### Technical Info
* PsPM revision ID: model inversion done in v5.0.0 (latest released version by 03.05.2021); model review done in PsPM GUI called from version "pspm-svn-r801-trunk".
* MATLAB version: 2018b

teddychao:
Fixes #213 .

1. Update the main GUI UI and the first level model review UI.
2. Update the displaying matrix to avoid confusing displaying caused by the NaN values.

File updates
- pspm
  - Some functions were just regenerated due to GUI update, but no modifications were added.
- pspm_rev_glm
  - Updated to UTF-8
  - Added a warning if the output is dominated by some outliers (L46-54)
  - Converted NaN to 0 for design matrix displaying only (L78)
- pspm_review
  - Some functions were just regenerated due to GUI update, but no modifications were added.
- pspm_ui and pspm_ui_installation
  - Only added some parameters required for the new UI in pspm_review


teddychao:
Hi @SerenXia 

I have investigated the issue. I think there may not be an issue of the code but a display setting... The result (Figure 1) we found homogeneous from x=1 to x=80 is due to the very different value at x=88 (around). If you check Figure 2 where the x axis is set to be between 1 and 80, there is visual diversity of the plot.
Figure 1
Figure 2
As the solution, I added some specification for displaying the plot that allow users to limit x axis to the region they want to observe in detail.

teddychao:
The main GUI and review model GUI have been updated as following
image
image
Features are not changed.

teddychao:
Hi, I have found the reason why the figure looks so different when displaying the first 88 columns. That is because the displaying matrix contains lots of nan values. When displaying the whole matrix, for the columns except for column 89, such nan values were displayed as black pixels, and the non-nan values were displayed as white pixels. This is why this [figure](https://user-images.githubusercontent.com/3895146/150662621-4d59a598-b3b1-42f7-affc-6dd444c6d877.png) showed so many lines. When displaying only the first 88 columns, both nan and very small values were displayed as black pixels, and some relatively larger values were displayed as white pixels. That is why the background looks like dark and a line can be visualised.
To resolve this issue, I manually set the nan values to be 0. This will help the displaying pixels to become what we expect in most cases.

dominikbach (approved):

PR #343, opened 2022-01-24
Fix issue 308 - missing flank

### LINKED ISSUE #308: pspm find marker events: flank setting
### Summary
optional field "flank" for marker events is only set available in pspm_get_marker.m and pspm_get_event.m, but not in pspm_import.m

When we import datafiles into pspm, we call function "pspm_import" to go through all data type and preprocess markers.
In pspm_import, line 201, depending on data type, specific function will be called. 

So for markers, function "pspm_get_marker" will be called, with "import" as input argument. 

In this function, function "pspm_get_events" will be then called, again with "import" as input argument.

In "pspm_get_events" and "pspm_get_marker", structure "import" contains an optional field "flank" to determine the orientation of marker calculation (see get events function). However, this was not available in pspm_import, meaning, although flank option was set, users actually have no way to use it.

### Technical Info
* PsPM revision ID: 5.1.1
* MATLAB version: 2018b

teddychao:
Fixes #308.

We discussed days ago about auto-filling some fields in header. We agreed that `pspm_load_data` should have this feature. Therefore I re-wrote `pspm_load_data` and left a section for auto-filling some fields in the future.

Changes proposed in this pull request
- General
  - All files are converted into UTF-8 and LF.
- New Features
  - `pspm_load_data`
    - The script has been re-written and tested functionally fully and compatible with the other functions.
    - The script now has a section 7 for autofilling parameters.
    - The description part was reformatted.
  - `pspm_get_marker`
    - Now the script detects and adds `flank` while applicable.
  - `pspm_import`
    - Now the script detects and adds `flank` while applicable.
- Test function update
  - `load_data_test` now uses a new compare function to compare `header` and `data`, because now `load_data` will auto-fill some fields.


dominikbach (changes requested):
1. flank should **not** be a field in the data structure. Please delete lines 40-43 of pspm_get_marker, and adapt section 7 of pspm_load_data.
2. Help text on flank should go into pspm_import (can just copy over from pspm_get_marker).

dominikbach (approved):

PR #344, opened 2022-01-24
Fix issue 212 - pspm_write_channel

### LINKED ISSUE #212: Allow empty marker channels
### Feature Description

Some functions assume that there is always data in every channel. However, it would make sense to have empty marker channels. 

To be changed: pspm_write_channel (line 86-116), pspm_load_data

teddychao:
Fixes #212.

- `pspm_write_channel`
  - use a 1-by-0 matrix for empty data field when writing results. Line 97--100
- `pspm_load_data`
  - use a 1-by-0 matrix for empty data field when writing results: Line 203


src/pspm_write_channel.m
```diff
@@ -95,17 +95,21 @@
...
-                    newdata{i}.data = d;
+                  warning('ID:invalid_data_structure', ['Passed struct (%i) seems to have the wrong ', ...
+                    'orientation. Trying to transpose...'], i);
+                  d = d';
+                  newdata{i}.data = d;
                 else
-                    warning('ID:invalid_data_structure', ...
-                        'Passed struct (%i) seems to have the wrong format.', i);
-                    return;
+                  warning('ID:invalid_data_structure', ...
```
teddychao, line 108: The `^M` is caused by the difference between CRLF and LF. I will update that after the approval of this pull request, because updating will highlight too many lines and confuse reviewers.

src/pspm_write_channel.m
```diff
@@ -95,17 +95,21 @@
         if isfield(newdata{i}, 'data') && isfield(newdata{i}, 'header')
             d = newdata{i}.data;
             [h,w] = size(d);
-            if w ~= 1
+            if h==0 && w==0
```
teddychao, line 98: As I understand, "the empty data field" indicates that "newdata{i} has a field data, but it is empty or 0x0", thus I add the update for this situation.

dominikbach (changes requested):
Seems like pspm_load_data might still through an error through a later check on the variable nflag?

teddychao:
> Seems like pspm_load_data might still through an error through a later check on the variable nflag?

Apologies I marked the wrong pull request as completed. It should be the pspm_scr_pp one which is completed, not this one. However I really need to finish this one today.

teddychao:
> Seems like pspm_load_data might still through an error through a later check on the variable nflag?

@dominikbach Sorry I am not sure which error `pspm_load1_data` brings. From the terminal I did see `pspm_load1_data` will report a warning since the data field is empty. I am not sure if that is expected. I think it may be useful to change it to size based warning. For example, there will be a warning if it is a 0x0 matrix, but there will not be a warning if it is 1x0, because 1x0 means there is no data in that field and the field has been regularised.

dominikbach (approved):

PR #352, opened 2022-02-28
Fix issue 351 - pspm_scr_pp

### LINKED ISSUE #351: Bug in writing channel (pspm_scr_pp)
### Summary

When writing the new SCR artefact channel, the header of the first channel of the data is copied (ligne 258). This is a problem when the first channel is not an SCR channel. 

I got very confused why the pspm_scr_pp added an empty respiration channel to my data. This is because the first channel of this dataset is a respiration channel and not a scr one. 

### Expected Results

It should not just copy the first channel's header. 

### Actual Results

It should be an SCR header. 

change ligne 258 of pspm_scr_pp

### Technical Info
* PsPM version: most recent
* MATLAB version: R2021b

teddychao:
Fixes #351 .

- `pspm_scr_pp`
  - Update to make `pspm_scr_pp` compatible with data whose first `chantype` is not `scr`.
- `pspm_scr_pp_test`
  - Update to have a test data file whose first channel is not `scr`. This is currently not working as it brings issue to `pspm_get_events`. Further investigation is required.

---
This pull request can be updated after the approval of `pspm_dcm` for checking incompatibility.


teddychao:
@jusporrer Thanks for pointing out the mistake. Meanwhile could you please let me know how this issue appears? As I understand, `pspm_scr_pp` should only handle `scr` data. I am not sure why the header could be another channel type?

jusporrer (changes requested):
I am afraid your solution is not enough because I just noted it's part of a bigger problem. 

When I call pspm_scr_pp, I input the file name of my data (which does not just contain a scr channel but also includes other channels).  This means you load all of these channels on line 168. This results in another problem on line 174 as the indata is from the first channel (which in my case, was not even a scr channel so the pp was not done on the scr). 

I think the best solution would be to follow the same format as pspm_resp_pp (include lines 47 to 51) and load only the relevant channel (like in line 75).

jusporrer:
I just committed a possible solution based on pspm_resp_pp. I think it should fix the problem. I have tested it in my dataset but not extensively.

teddychao:
> When I call pspm_scr_pp, I input the file name of my data (which does not just contain a scr channel but also includes other channels). 

Thanks for the solution. May I ask why `pspm_scr_pp` is called when there is not a `scr` channel to process? I agree the loading function shall just load the relevant channel instead of all data. I am asking this because I have to update the test function to avoid relevant issues appearing in the future. As I understand, it seems like a datafile which does not have a `scr` channel is imported to and processed by `pspm_scr_pp`. If this is a situation that users may need to do, I will add a test case in its test function.

jusporrer:
Sorry if this was not clear but there was definitely a scr channel to process. The only difference was that the scr channel was not in the first channel (which was usually the case for me previously).  I don't think it's necessary to accommodate situations in which a datafile that does not have a scr channel is imported to and processed by pspm_scr_pp.

jusporrer:
Did you see my commit? It should address this :)

jusporrer:
No worries, thank you for your help!

teddychao:
I don't quite understand where the problem is. I tested on my computer with `pspm_github_actions` and `pspm_test`, all passed perfectly.
image
However, I got this issue when using GitHub Actions
image
I did not see the correlation between `pspm_get_events_test` and `pspm_scr_pp_test`. I think Juliana's solutions works perfectly but I don't know why there is a rise issue in `pspm_get_events_test` when I added
```
channels{1}.chantype = 'hb';
channels{2}.chantype = 'scr';
scr_pp_test_template(this, channels)
```
Currently I disabled these three lines as they work on my computer but not on GitHub Actions. Users may need the updated version of `pspm_scr_pp` thus I think it is better to include first. I may need to propose another issue to find why `pspm_get_events_test` is affected by `pspm_scr_pp_test` in our situation.

jbrochar (approved):
It looks good for me, from a code point of view.
I cannot say anything from the functionality point of view.

dominikbach (approved):

teddychao:
Thanks for approving the changes. I cannot merge it because it told me changes have not been addressed. I am not sure how to address it or I have to ask Juliana to approve..

dominikbach (approved):

Issue #248, opened 2021-06-17
Data Display "Semi automatic mode" in Pre-processing of Heart data


Wjelena:
### Summary
When preprocessing heart data with semi automatic mode "on", the data will sometimes not be properly displayed in the ecg editor, but instead in the PsPM GUI, or Batch.

### Expected Results
![ecg_editor](https://user-images.githubusercontent.com/47207904/122422221-459a2380-cf8d-11eb-92b0-327a9d3fc8ab.JPG)

### Actual Results
![batch_issue_ecg_editor](https://user-images.githubusercontent.com/47207904/122422240-492daa80-cf8d-11eb-84d4-b8ed4de4ddb7.JPG)
![batch_issue_ecg_semi](https://user-images.githubusercontent.com/47207904/122422264-4d59c800-cf8d-11eb-9b2a-8b4c95796f0c.JPG)

### Technical Info
* PsPM revision ID: 5.1.0 beta
* MATLAB version: 2018b

teddychao:
@Wjelena Could you please send me the sample data so I can reproduce the error. Thanks.

Wjelena:
hi Teddy

it's the DoxMem2 dataset that you also used for yanfang yesterday. Attached you can find a file to test it.

Best wishes
Jelena

--
Jelena Wehrli, MSc
Psychiatric University Hospital
***@***.***
bachlab.org
+41 44 384 26 60

-----"Teddy" ***@***.***> schrieb: -----
An: "bachlab/PsPM" ***@***.***>
Von: "Teddy" ***@***.***>
Datum: 17.06.2021 17:02
Kopie: "Wjelena" ***@***.***>, "Mention" ***@***.***>
Betreff: Re: [bachlab/PsPM] Data Display "Semi automatic mode" in Pre-processing of Heart data (#248)


 @Wjelena Could you please send me the sample data so I can reproduce the error. Thanks.

teddychao:
Hi Jelena,

Thanks for the data. I am not sure which operation leads to this error. In general I guess it could be cause by the GUI. Could I know if you remember which operation leads to that issue? for example when you zoom in or zoom out the figure (I tried just now but I did not see error when I zoomed in or out).
I think the ECG editor GUI will be improved in the next release, so I will keep an eye on it.

Best wishes
teddy

Wjelena:
Hi Teddy

Thank you for the reply! I tried a few more times and could not replicate it now, so I am not sure what I did before to cause this.
[...]

PR #359, opened 2022-03-16
Fix Issue 358 - pspm_data_editor

### LINKED ISSUE #358: pspm_data_editor is not working
### Summary
pspm_data_editor is not working. When users import a datafile, pspm_data_editor does not display the plot appropriately.

### Steps to Reproduce
1. Open pspm_data_editor.
2. Import a datafile.

### Expected Results
The results should be displayed at the middle of the window.

### Actual Results
The results are not displayed, while a new plot window is brought to users.

### Technical Info
* PsPM version: develop
* MATLAB version: 2021b

teddychao:
Fixes #358.

- Update `pspm_data_editor` UI.
- Update targeting plot towards axData rather than the current figure (GUI).


teddychao:
@dominikbach Thanks for posting the issue. The issue is caused by some code which assumes that the data_editor figure (GUI) is always able to be called and the current figure. Due to a recent issue of incorrect plotting ([Issue 248](https://github.com/bachlab/PsPM/issues/248)), we changed all the GUI files to be inaccessible by callback functions. However, the code at line 315 assumes the current data_editor GUI can be called. When the data_editor GUI cannot be called, it generates a new window, which is what you see there. I think it needs some work tomorrow. I have temporarily set the GUI file accessible to callback functions.

teddychao:
This issue has been resolved. However I think there are more issues. For example the two buttons and right bottom corner seem to be weird. When I close the GUI by clicking 'Cancel' and open data_editor again, I will not be able to load new data again. I don't know why but that is another issue. The plot is currently hard to read since the text in the figure is very small and the measurement (unit) for the X-Y axis is missing. I think it requires another pull request, but the current version has been functional and solved the current issue.
image
image

dominikbach (approved):
Good to see it's working now - and please do open new issues for the observed problems.

PR #363, opened 2022-04-11
Fix issue 361 - pspm_dcm

### LINKED ISSUE #361: pspm_dcm throw error of inconsistent matrix when processing missing_epochs
### Summary
`pspm_dcm` reports error when processing with `missing_epochs_error`. The error showed like 
> Error in pspm_dcm (line 570)
> sbs_miniti(sbs_id) = min(sbs_iti{sbs_id}(1 : end - 1));
> Unable to perform assignment because the left and right sides have a different number of elements.
The error did not appear when processing certain time epochs.
The issue could be due to inconsistent information obtained or processed for missing_epochs_error in comparison with the object. Users have provided data for reproduction.

### Technical Info
* PsPM version: develop

teddychao:
Fixes #361.

The reason that leads to the issue:
- In `pspm_dcm`, at line 616 (original version), `proc_miniti` was updated by removing NaN. This makes the length of `proc_miniti` equal or smaller than the original size of `proc_miniti` or the size of `sbs_miniti`.
- In line 706 -- 707, `proc_miniti` was processed together with `isbSn`, but `isbSn` was determined based on `model.scr`, and `model.scr` was set consistent to `sbs_miniti`. This leads to the bug.

- General
  - `pspm_test` and `pspm_test_github_actions` now have included the tests for `pspm_dcm`.
- Bug fix
  - `pspm_dcm`
    - Some variables were not updated properly (see comment), which are now updated reasonably by considering how they are processed in the following processing.


src/pspm_dcm.m
```diff
@@ -130,34 +154,31 @@
...
-
 % cell array which saves all the warnings which are not followed
 % by a `return` function
 warnings = {};
 
-%% Check input arguments & set defaults
-% -------------------------------------------------------------------------
+%% 2 Check input arguments & set defaults
 if nargin < 1
-  warning('ID:invalid_input', 'No data to work on.'); fn = []; return;
```
teddychao, line 149: I think the script has been terminated in this case and no outputs will be provided... thus we would not need to redefine fn again here.

dominikbach (approved):

jbrochar (approved):
Nothing to add/comment, this looks good to me.
(from a coding point of view)

PR #364, opened 2022-04-11
Fix issue 302 - test mode for PsPM

### LINKED ISSUE #302: Test mode for PsPM
### Feature Description

Option to allow silent or simplified running report in Terminal

Currently PsPM will try to export the essential processing report in MATLAB terminal. Sometimes, users may be just interested in whether the processing completes successfully instead of knowing the details, especially for those who use GUI. Therefore, is it possible to add an option in PsPM, which allows to silent or simplify the terminal report?

teddychao:
Fixes #302.

- Update to provide an option of develop mode
- Add a new function `pspm_overwrite` to handle overwriting operations.

Please kindly keep this pull request open after approving it because I need to sort out some formatting issues and documents that I have not finished. Thanks.

### DIFF
--- a/src/backroom/pspm_find_data_epochs.m
+++ b/src/backroom/pspm_find_data_epochs.m
@@ -99,12 +99,28 @@ end;
 % create epochs and divide by samplerate
 epochs = ([start, stop]).*data{chan_id}.header.sr^-1;
 
+file_exist = exist(options.output_file, 'file');
+write_ok = false;
+if file_exist
+    if options.overwrite 
+        write_ok = true;
+    elseif ~options.dont_ask_overwrite
+        if feature('ShowFigureWindows')
+            msg = ['File already exists. Overwrite?', newline, 'Existing file: ',options.output_file];
+    	    ov = questdlg(msg, 'File already exists', 'Yes', 'No', 'Yes'); % default to overwrite by users
+        else
+            ov = 'Yes'; % default to overwrite on Jenkins
+        end
+        write_ok = strcmp(ov,'Yes');
+    end;
+else
+    write_ok = true;
+end;
 
-if pspm_overwrite(options.output_file, options)
+if write_ok
     save(options.output_file, 'epochs');
 end;
 
 epochfile = options.output_file;
 
 sts = 1;
-
--- a/src/backroom/pspm_find_data_epochs.m
+++ b/src/backroom/pspm_find_data_epochs.m
@@ -99,24 +99,8 @@ end;
 % create epochs and divide by samplerate
 epochs = ([start, stop]).*data{chan_id}.header.sr^-1;
 
-file_exist = exist(options.output_file, 'file');
-write_ok = false;
-if file_exist
-    if options.overwrite 
-        write_ok = true;
-    elseif ~options.dont_ask_overwrite
-        if feature('ShowFigureWindows')
-            msg = ['File already exists. Overwrite?', newline, 'Existing file: ',options.output_file];
-    	    ov = questdlg(msg, 'File already exists', 'Yes', 'No', 'Yes'); % default to overwrite by users
-        else
-            ov = 'Yes'; % default to overwrite on Jenkins
-        end
-        write_ok = strcmp(ov,'Yes');
-    end;
-else
-    write_ok = true;
-end;
 
+write_ok = pspm_overwrite(options.output_file, options);
 if write_ok
     save(options.output_file, 'epochs');
 end;
--- a/src/backroom/pspm_find_data_epochs.m
+++ b/src/backroom/pspm_find_data_epochs.m
@@ -100,8 +100,7 @@ end;
 epochs = ([start, stop]).*data{chan_id}.header.sr^-1;
 
 
-write_ok = pspm_overwrite(options.output_file, options);
-if write_ok
+if pspm_overwrite(options.output_file, options)
     save(options.output_file, 'epochs');
 end;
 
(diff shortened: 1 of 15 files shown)


src/backroom/pspm_scr2ledalab.m
```diff
@@ -101,23 +101,15 @@
...
-    end             
-    close gcf;
-else
-    overwrite = 'Yes';
-end;
-if strcmp(overwrite, 'No')
-    warning('Data discarded ...');
-    return;
+
+if pspm_overwrite(outfile, options)
```
jbrochar, line 105: [minor comment]
For consistency, shouldn't the syntax be:
```
write_ok = pspm_overwrite(options.output_file, options);
if write_ok
```
as in src/backroom/pspm_find_data_epochs.m line 103?

src/backroom/pspm_scr2ledalab.m
```diff
@@ -101,23 +101,15 @@
...
+	save(outfile, 'data', 'fileinfo');
 else
-    save(outfile, 'data', 'fileinfo');
-end;
+	warning('Data discarded ...');
```
jbrochar, line 108: [minor comment]
Would a naive user reading this warning be able to understand where it comes from and why?
Could it be extended with more information, while still being easy to read?

src/pspm_overwrite.m
```diff
@@ -0,0 +1,89 @@
...
+% pspm_overwrite generalises the overwriting operation
+% pspm_overwrite considers the following situations
+% - whether options.overwrite is defined
+% - whether PsPM is in develop mode
+% - whether the file exist
+% - whether GUI can be used
+% Only the following situation will stop PsPM to overwrite files
+% 1. if overwrite is defined as not to overwrite
+% 2. if overwrite is not defined, and users use the GUI to stop overwriting
+
```
jbrochar, line 13: Wouldn't this blank line (without a `%` at the beginning) prevent the documentation to be correctly generated?
With and without a `%` it gives me a full or a truncated help section.
image
image

src/pspm_overwrite.m
```diff
@@ -0,0 +1,89 @@
...
+% ow        option of overwrite if this option is presented
+%           can be a value or a struct
+%           if a value, can be 0 (not to overwrite) or 1 (to overwrite)
+%           if a struct, check if the field "overwrite" exist
+
+% OUTPUTS
+% ow_final  option of overwriting determined by pspm_overwrite
+
+pth = fileparts(which('pspm'));
+settings_struct = load([pth, '/pspm_settings.mat'], "settings");
```
jbrochar, line 26: [minor comment]
Isn't it time-consuming to load the settings like that each time, instead of using a global variable?
It is a genuine question, it might not worth the effort to change this way of doing.

src/pspm_overwrite.m
```diff
@@ -0,0 +1,89 @@
...
+    return;
+  case 1 % ow is not defined
+    fn = varargin{1};
+    if iscell(fn)
+      fn = fn{1};
+    end
+    switch settings.developmode
+      case 1
+        ow_final = 1; % In develop mode, always overwrite
+      otherwise
```
jbrochar, line 44: It is a bit confusing to use a switch for a single case.
Are you planning to add multiple `settings.developmode` in the future?
If so, please discard this comment.

src/backroom/pspm_scr2ledalab.m
dominikbach: Just to clarify: "Backroom" includes functions that cannot be accessed from the GUI - this is for development and legacy purposes.

dominikbach (commented):
Just two generic comments:
- pspm_overwrite seems a very good idea
- in general I'd suggest not touching backroom functions (even if incompatible with the main code) as they are kept for legacy and development purposes only

jbrochar (approved):
There are still a few minor comments of mine here and there, but I'll let Teddy judge if they should lead to code change or not, none of them are critical.

src/backroom/pspm_scr2ledalab.m
teddychao: > [minor comment]
> Would a naive user reading this warning be able to understand where it comes from and why?
> Could it be extended with more information, while still being easy to read?

Yes.. Some warnings caused by overwriting are proposed in `pspm_overwrite` thus they are not repeated here. I am not the author of this script so I have not done any updates in this function but just applied the new APIs.

src/pspm_overwrite.m
teddychao: Yes... I have deleted that blank line break.

src/backroom/pspm_find_data_epochs.m
```diff
@@ -99,24 +99,8 @@
...
-            ov = 'Yes'; % default to overwrite on Jenkins
-        end
-        write_ok = strcmp(ov,'Yes');
-    end;
-else
-    write_ok = true;
-end;
 
+write_ok = pspm_overwrite(options.output_file, options);
 if write_ok
```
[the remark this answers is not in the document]
teddychao: I agree this should be consistent throughout these functions. I tried to make it clear as the output of pspm_overwrite can be found in its help text. Producing less variables may make the software run faster.

src/pspm_overwrite.m
teddychao: > It is a bit confusing to use a switch for a single case.
> Are you planning to add multiple `settings.developmode` in the future?
> If so, please discard this comment.

I wrote it like that because `settings.developmode` may be defined as something else and I want to avoid such bugs... However I agree it is a little confusing. I have changed it to logical values now.

teddychao:
> Just two generic comments:
> 
> * pspm_overwrite seems a very good idea
> * in general I'd suggest not touching backroom functions (even if incompatible with the main code) as they are kept for legacy and development purposes only

No problem, thanks for reminding. I have reverted the backroom functions.

PR #378, opened 2022-05-09
Re-fix issue 375 - pspm_data_editor

### LINKED ISSUE #375: data_editor not displaying missing epochs when being run from command line
### Summary
1. `pspm_data_editor` is expected to plot the missing epochs when they are loaded from either the command line or the GUI, however currently only the latter option is working.
2. `pspm_data_editor` does not recognise the overwrite options specified in the command line version.
### Technical Info
* PsPM version: develop
* MATLAB version: 2022a

teddychao:
Re-fix #375

This is a patch for re-fixing issue 375.
The error, showing as the vanished missing epoch block in the GUI and the un-annotated signal, is caused by incorrect logic of plotting data and adding epochs.

- Update pspm_data_editor.m


dominikbach (approved):
Tested in on my data and it seems to work. Thanks very much!

PR #394, opened 2022-05-31
Fix issue 224 - pspm_data_editor

### LINKED ISSUE #224: Data editor to show missing epochs in time order
### Feature Description
Data editor is currently showing missing epochs based on the default order. Users may prefer to see the missing epochs listed in the time order.

teddychao:
Fixes #224 .

- The epoch list in `pspm_data_editor` now sorts the epoch list based on the time order.
- The plots in `pspm_data_editor` now display the units for the x and y axises.


src/pspm_data_editor.m
```diff
@@ -838,7 +841,13 @@ function UpdateEpochList(hObject)
...
   end
   set(handles.lbEpochs, 'String', names);
-  handles.epochs = epochs;
+  % update to sort epochs according to the time series order
+  epochs_list = 1:length(epochs);
+  for i_epochs_list = 1:length(epochs) % obtain data
+    epochs_list(i_epochs_list) = epochs{1,i_epochs_list}.range(1);
+  end
+  [~,idx]=sort(epochs_list);
+  handles.epochs = epochs{1,idx};
```
jbrochar, line 850: FYI I think the six previous lines can be shorten as:
`handles.epochs = sort( cellfun(@(epoch) epoch.range(1), epochs) )`
Which might be less intuitive to read or not. I let you be the judge of that

jbrochar (commented):
Good from a coding perspective.
I added a minor suggestion.

src/pspm_data_editor.m
teddychao: Great idea, thanks Jules!

jbrochar (approved):

src/pspm_data_editor.m
jbrochar: Oh? There might be a typo from my end then.
What is the output of `cellfun(@(epoch) epoch.range(1), epochs)`?
Alternatively, can you send me be email (jules.brochard@gmail.com) an example of the variable `epochs`?

teddychao: > Oh? There might be a typo from my end then.
> What is the output of `cellfun(@(epoch) epoch.range(1), epochs)`?
> Alternatively, can you send me be email ([jules.brochard@gmail.com](mailto:jules.brochard@gmail.com)) an example of the variable `epochs`?

No problem, here are two screenshots showing the results... I will send you the link to the test files soon.
original solution
Screenshot 2022-06-03 at 16 48 17
proposed solution
Screenshot 2022-06-03 at 16 47 28

jbrochar: it is strange indeed.

What is the output of `cellfun(@(epoch) epoch.range(1), epochs)`? 
This should simply extract the .range(1) field of every element of the cell array `epochs`.

Could you also try `cellfun(@(epoch) epoch.range(1), epochs(1,:))` ?

PR #398, opened 2022-06-14
Fix issue 395 - pspm_con1

### LINKED ISSUE #395: Incorrect assigned NaN
### Summary
If there is a single NaN in `dcm.stats`, `pspm_con1` will assign `NaN`s to every single `dcm.con` field.
### Steps to Reproduce
Process with the following code
```
convec = [1/10, 0, -1/6, 0, 1/10, 0, -1/6, 0, 1/10, 0, 0, -1/6, 1/10,
0, 1/10, 0, -1/6, 0, 1/10, 0, 0, -1/6, 0, 1/10, 0, -1/6, 0, 1/10, 0,
1/10, 1/10, 0]
pspm_con1('modelfile_with_nan.mat', {'CS+US_vs_CS+noUS'}, {convec},
'param', 1);
```
### Expected Results
`-0.0214 0.1689 0.1643 -0.0773 0.0059 0.1615`
### Actual Results
`NaN NaN NaN NaN NaN NaN`
### Technical Info
* PsPM version: develop
* MATLAB version: 2022a

teddychao:
Fixes #395.

- `pspm_con1`
  - Update the method of calculating `conval`.

#### Explanation
Definitions

- **A1**. Define Line *N* does not have issue if data.stats(N,:) does not have NaN.
- **A2**. Define Line *N* has a valid issue if data.stats(N,:) does have NaN and conmat(N)==0.
- **A3**. Define Line *N* has an invalid issue if data.stats(N,:) does have NaN and conmat(N)~=0.

Therefore, there are 3 situations when there are issue(s)

- **B1**. There are issue(s), all issues are valid.
- **B2**. There are issue(s), all issues are invalid.
- **B3**. There are issues, issues are valid and invalid.

Among the three situations, **B2** and **B3** will lead to the same results of conval as `NaN`, thus the processing for **B2** and **B3** is the same.
For **B1**, the valid parts are converted to be 0 for calculating conval.
The logic is therefore as

> Check if data.stats has `NaN`
> ↓
> If there are `NaN`s, there are issues, and the issue(s) could be **B1**, **B2** or **B3**. **B2** and **B3** lead to the same processing.
> ↓
> For **B1**, convert the NaN rows to be 0 in `data.stats`, then calculate `conval`. Give a warning "Calculated data.stats contain NaN caused by 0 in conmat, which are ignored when calculating conval".
> For **B2** and **B3**, calculate `conval` as `NaN`. Give a warning "Calculated data.stats contain `NaN`s, caused by unknown reason, which is invalid."


src/pspm_con1.m
```diff
@@ -1,176 +1,164 @@
...
+  % 3.1 user output --
   fprintf('Loading data ... ');
-
-  % retrieve stats --
-  [sts, data, mdltype] = pspm_load1(modelfile{iFn}, datatype, '', load1_options);
-  if sts == -1
+  % 3.2 retrieve stats --
+  [lsts, data, ~] = pspm_load1(modelfile{iFn}, datatype, '', load1_options);
+  if lsts == -1
     warning('ID:invalid_input', 'Could not retrieve stats');
```
jbrochar, line 96: Simple warning or actual error?

src/pspm_con1.m
```diff
@@ -1,176 +1,164 @@
...
+    end
+  end
+  % 3.4 user output --
   fprintf('\nWriting contrasts to %s\n', modelfile{iFn});
-  % check number of contrast weights --
+  % 3.5 check number of contrast weights --
   paramno = size(data.stats, 1);
   for c=1:numel(convec)
     if numel(convec{c}) > paramno
       warning('Contrast (%d) has more weights than statistics (%d) in modelfile %s', ...
```
jbrochar, line 117: warning or error?

src/pspm_con1.m
```diff
@@ -1,176 +1,164 @@
...
-
-  % calculate contrasts --
+  end
+  % 3.7 calculate contrasts
   % this automatically replicates contrasts across multiple measures if
   % data.stats has more than one column
-  conval = conmat * data.stats;
-
+  if sum(isnan(data.stats)) > 0
+    warning('Calculated statistics contain NaN, save the results?');
```
jbrochar, line 131: warning or error?

jbrochar (changes requested):
Same point than in the other review, I prefer actual errors to warnings (but it's not my call in the end :) )

Dominik said to avoid throwing errors, so don't mind my comments :)

src/pspm_con1.m
```diff
@@ -1,176 +1,176 @@
...
-  % calculate contrasts --
+        numel(convec{c}), paramno, modelfile{iFn});
+      return
+    end
+  end
+  % 3.6 transform into row vector and right pad with zeros --
+  conmat = zeros(numel(connames), paramno);
+  for c = 1:numel(convec)
+    conmat(c,1:numel(convec{c})) = convec{c};
+  end
```
jbrochar, line 126: Tip: If all cells of `convec `are row-vector, I think `conmat = cat(1,convec{:});` is similar to your for loop

src/pspm_con1.m
```diff
@@ -1,176 +1,176 @@
...
   % data.stats has more than one column
-  conval = conmat * data.stats;
-
+  % there are issues if data.stats has NaN and the corresponding conmat
+  % also contains 0
+  idx_issue = isnan(prod(data.stats,2));
+  idx_valid_issue = transpose(conmat==0) .* idx_issue;
+  idx_invalid_issue = transpose(conmat~=0) .* idx_issue;
+  if sum(idx_issue) > 0
+    if sum(idx_valid_issue) > 0 && sum(idx_invalid_issue) == 0
```
jbrochar, line 136: Here, the `sum `are run on matrices, right?
If so, then it is a bit confusing if you run an `if `on the resulting boolean **vector**

jbrochar (changes requested):
The current code won't handle multiple contrasts.

teddychao:
Hi @jbrochar Thanks for discussing with me about this issue today. Could you please try this data (just load it in
Screenshot 2022-06-28 at 17 25 39
 matlab) and then run the code in this script (maybe just run them in terminal)? I think it can run and it contains two rows of NaNs? thanks
[Archive.zip](https://github.com/bachlab/PsPM/files/9002935/Archive.zip)
.

src/pspm_con1.m
```diff
@@ -1,176 +1,176 @@
...
+    conmat(c,1:numel(convec{c})) = convec{c};
+  end
+  % 3.7 calculate contrasts
   % this automatically replicates contrasts across multiple measures if
   % data.stats has more than one column
-  conval = conmat * data.stats;
-
+  % there are issues if data.stats has NaN and the corresponding conmat
+  % also contains 0
+  idx_issue = isnan(prod(data.stats,2));
```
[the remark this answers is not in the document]
teddychao: Yes, you solution has a higher readability. I have updated the code in this line 🙂

src/pspm_con1.m
```diff
@@ -1,176 +1,176 @@
...
+  idx_valid_issue = transpose(conmat==0) .* idx_issue;
+  idx_invalid_issue = transpose(conmat~=0) .* idx_issue;
+  if sum(idx_issue) > 0
```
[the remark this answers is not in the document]
teddychao: Yes! I have changed both of this line and the following line to replace `sum` by `any` now.

src/pspm_con1.m
teddychao: The sum here is just used to detect if there are positive values. I have changed it to `any` as they are really Boolean vectors. In the new version, these operations are done in every line thus `idx_valid_issue` and `idx_invalid_issue` are both vectors.

src/pspm_con1.m
teddychao: This is not written by me. I just checked and I think they do have similar features. I have updated the code based on your suggestion. Now sure if we need to ask Dominik to check as he is the author of this line of code.

jbrochar (approved):

PR #401, opened 2022-07-12
Re-fix issue 395

### LINKED ISSUE #395: Incorrect assigned NaN

teddychao:
Re-Fixes #395 .

- Update pspm_con1.m
  - In the previous version, `conval = conmat * data_stats_converted` will be executed only if there are detected NaNs and will be done in every loop, which is not expected. Currently, `conval = conmat * data_stats_converted` is done in the final loop. The sizes of these variables are
    - `conmat` $a\times b$
    - `data_stats_converted` $b\times c$.
    - `conval` $a\times c$.


src/pspm_con1.m
```diff
@@ -141,14 +141,14 @@ function pspm_con1(modelfile, connames, convec, datatype, deletecon, options)
...
-        conval = conmat * data_stats_converted;
       else
         warning(['Calculated data.stats contain NaNs that are caused by unknown reasons. '...
           'However they were not used in the computation of the contrasts.']);
-        conval = conmat * data.stats;
+        data_stats_converted = data.stats;
       end
     end
   end
+  conval = conmat * data_stats_converted;
```
jbrochar, line 151: No, this does not work, because data_stats_converted is rewritten at every iteration before the contrast value is actually written.
I suggest you focus on your deadline, and we arrange a meeting to come back to this issue together. In the meantime the release should be suspended.

jbrochar (changes requested):

src/pspm_con1.m
```diff
@@ -130,26 +130,27 @@ function pspm_con1(modelfile, connames, convec, datatype, deletecon, options)
...
       idx_valid_issue = transpose(conmat_array==0) .* idx_issue;
       idx_invalid_issue = transpose(conmat_array~=0) .* idx_issue;
-      if any(idx_valid_issue) && ~any(idx_invalid_issue)
-        warning(['Calculated data.stats contain NaNs that are caused by unknown reasons. '...
-          'data.stats are then used in a contrasts, producing an invalid results (NaN).']);
-        data_stats_converted = data.stats;
-        data_stats_converted(isnan(data_stats_converted)) = 0;
-        conval = conmat * data_stats_converted;
-      else
+      if any(idx_valid_issue)
```
jbrochar: We should be careful here, the NaN are not a problem if and only if there are no `invalid issue`.
Shouldn't the text be `if any(idx_invalid_issue)` ?

src/pspm_con1.m
```diff
@@ -130,26 +130,27 @@ function pspm_con1(modelfile, connames, convec, datatype, deletecon, options)
...
         warning(['Calculated data.stats contain NaNs that are caused by unknown reasons. '...
           'However they were not used in the computation of the contrasts.']);
-        conval = conmat * data.stats;
+        data_stats_converted_array = data_stats_converted(idx_valid_issue==1,:);
+        data_stats_converted_array(isnan(data_stats_converted_array)) = 0;
+        data_stats_converted(idx_valid_issue==1,:) = data_stats_converted_array;
```
jbrochar: ? I am a bit confused here. Isn't it equivalent to: `data_stats_converted(idx_valid_issue,:) = 0` ?

src/pspm_con1.m
```diff
@@ -130,26 +130,27 @@ function pspm_con1(modelfile, connames, convec, datatype, deletecon, options)
...
+      elseif any(idx_invalid_issue)
+        warning(['Calculated data.stats contain NaNs that are caused by unknown reasons. '...
+          'data.stats are then used in a contrasts, producing an invalid result (NaN).']);
       end
     end
   end
-
+  conval = conmat_converted * data_stats_converted;
```
jbrochar: I don't think this will provide the correct response for multiple contrast.
It seems to me that data_stats_converted keeps getting rewritten at each contrast.

For instance, if the contrast are [ 0, 1] and [1, 0] and the data have NaN in the first line.
With the current code the NaN will be converted to zero by the processing of the first contrast, and the computation of the second contrast will then not produce the NaN it ought to produce.

jbrochar (changes requested):

src/pspm_con1.m
teddychao: this is the old version.. has been updated

src/pspm_con1.m
teddychao: This is the old version and has been removed, but thanks for reminding me of it 🙂

teddychao:
> Beside a small change, I think this version is great!

Thanks! just can I know which small change you want me to add?...

jbrochar:
> > Beside a small change, I think this version is great!
> 
> Thanks! just can I know which small change you want me to add?...

The switch from 'any' to 'all'.

teddychao:
> Minor comment:
> isn't conval a column-vector? If that is the case, `conval(rconval) `should be enough (instead of `conval(rconval,:)`)

Do you mean if `conval` is a $1\times N$ matrix ($N>1$)? I do not think so, because we previously discussed about it and we found that `connames` (the argument of this function) can support multiple column names. In this case `conval` becomes a $M\times N$ matrix ($M, N>1$), that's why I used `conval(rconval,:)` here.

teddychao:
> > > Beside a small change, I think this version is great!
> > 
> > 
> > Thanks! just can I know which small change you want me to add?...
> 
> The switch from 'any' to 'all'.

I see, thanks for reminding though

dominikbach:
Looks good.

src/pspm_con1.m
```diff
@@ -129,27 +129,24 @@ function pspm_con1(modelfile, connames, convec, datatype, deletecon, options)
...
+        conval(rconval,:) = conmat(rconval,~IdxIssueRow) * data.stats(~IdxIssueRow,:);
         warning(['Calculated data.stats contain NaNs that are caused by unknown reasons. '...
-          'data.stats are then used in a contrasts, producing an invalid results (NaN).']);
-        data_stats_converted = data.stats;
-        data_stats_converted(isnan(data_stats_converted)) = 0;
-        conval = conmat * data_stats_converted;
+          'However they were not used in the computation of the contrasts.']);
       else
+        % the issues refer to non-0 or 0 in conmat
+        conval(rconval,:) = conmat(rconval,:) * data.stats;
```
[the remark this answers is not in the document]
teddychao: Sure, I have removed these two lines.

PR #438, opened 2022-12-11
Fix issue 437 - quote marker

### LINKED ISSUE #437: Quotations are used inconsistently
### Summary

Looking at `pspm_load_data`, quotation marks ('' and "") are used inconsistently.

Because they have a different meaning in more recent matlab versions, that has a potential to lead to errors.

PsPM should only ever use the single quotations ('').

teddychao:
Fixes #437.

- `pspm`, `src/pspm_load_data.m`
  - Replace `"` with `'`
- `src/pspm_options.m`
  - Replace `"` with `char(34)`
- `pspm_contrast`, `pspm_convert_gaze_distance`, `src/pspm_data_editor.m`, `src/pspm_dcm.m`, `src/pspm_extract_segments.m`, `src/pspm_filtfilt.m`, `src/pspm_get_events.m`, `src/pspm_get_labchartmat_in.m`, `src/pspm_get_timing.m`, `src/pspm_get_viewpoint.m`, `src/pspm_glm.m`, `src/pspm_overwrite.m`, `src/pspm_prepdata.m`, `src/pspm_trim.m`
  - Replace **"** with **`**
- `src/pspm_get_labchart.m`, `src/pspm_jobman.m`
  - Remove **"**
- `pspm_convert_ppg2hb`
  - `chan` is no longer used, so replaced with `channel`


dominikbach:
@teddychao This is good - but it would be sufficient to avoid the symbol " in the code, not necessarily in comments or within printed text (e.g. in warnings or sprintf)

dominikbach (approved):

teddychao:
> @teddychao This is good - but it would be sufficient to avoid the symbol " in the code, not necessarily in comments or within printed text (e.g. in warnings or sprintf)

Thanks @dominikbach. I was just trying to avoid using `"` in any situation, so that it can be easily detected which functions are still using " in the code by performing a global search in the folder. If `"` has to be used in some warning text, `char(34)` can be used in `print()` or `warning` currently.

PR #443, opened 2022-12-20
Fix issue 428 -- compute_visual_angle_core

### LINKED ISSUE #428: Operation `interpolate` in `pspm_compute_visual_angle_core`
### Summary
The action `interpolate`, currently defined in `pspm_compute_visual_angle_core`, does not seems to be very reasonable. It may need to be removed, and proper test for the output using `NaN` as the input is also required.

This issue can be resolved after merging the pull request #399.

teddychao:
Fixes #428.

From what I can check from the code, `options.interpolate` from `pspm_compute_visual_angle` was used in `pspm_compute_visual_angle_core` at line 36--42. Furthermore, extrapolation is computed in line 234 at `pspm_interpolate`. I think this is mainly used for filling the NaNs at the beginning and the end only instead of the middle... 

- Update `pspm_compute_visual_angle_core`.


teddychao:
If the interpolation action is removed in `pspm_compute_visual_angle_core`, there will be an error in `pspm_convert_visangle2sps_core` at line 12, because the code `says cannot calculate sps from data with NaN values`.

dominikbach:
> If the `interpolation` action is removed in `pspm_compute_visual_angle_core`, there will be an error in `pspm_convert_visangle2sps_core` at line `12`, because the code `says cannot calculate sps from data with NaN values`.

In my version of the function, line 12 is a comment (help text). Could you give more details where this error comes from?

dominikbach (approved):

PR #451, opened 2023-01-17
Fix issue 427 -- glm, merge, and extract_segments

### LINKED ISSUE #427: `marker_chan` / `marker_chan_num` unclearly defined
### Feature Description
`marker_chan` / `marker_chan_num`, used in `pspm_extract_segments`, `pspm_glm` and `pspm_merge`, are not clearly defined. Proper clarification is required for these two fields.

It is also useful to know whether `marker_chan` / `marker_chan_num` can be generalised into one field.

marker_chan | pspm_extract_segments | ? | ? | Mandatory if timeunit is 'markers'. For the function to find the appropriate timing of the specified marker ids. Must have the same format as data_fn. If timeunit is 'markers' and raw data are given then this parameter should be an cell array of numeric array of marker data.
-- | -- | -- | -- | --
marker_chan_num | pspm_glm | ? | the last marker channel number | double
marker_chan_num | pspm_merge | vector | [0 0] | any vector with two numeric values?

teddychao:
Fixes #427.

Updates
- `pspm_glm`, `pspm_sf`, `pspm_trim`
  - The default value of `marker_chan_num` needs to refer to the first channel.
- `pspm_merge`
  - The default value is `[1,2]`, meaning the first channel will be merged as the default option.
- `pspm_extract_segments`  
  - The first channel is always referred by `marker_chan` as default.
  - `marker_chan` is always converted into a cell if it is not a cell, and the size is based on the input data.


dominikbach:
From what I can see, the proposed solution will not yield the desired behaviour. 

1. The default marker channel should always be a *marker* channel, but it seems that the default is set to 1, i.e. the first channel of the file. This could be any channel.
2. Still, if there are several marker channels, the last channel will be selected e.g. in `events{iFile} = data{end}.data(:) * data{end}.header.sr;`

In my view, the correct option is:
1. Default argument for `marker_chan_num` should be `'marker'`; thus the function `pspm_load_data` will only load data channels.
2. Whether the first or the last marker channel is selected should be set in `pspm_glm` by using the first or the last element of the cell array returned by `pspm_load_data` .

dominikbach (changes requested):
Please see my previous comments.

teddychao:
> 1. Default argument for `marker_chan_num` should be `'marker'`; thus the function `pspm_load_data` will only load data channels.
> 2. Whether the first or the last marker channel is selected should be set in `pspm_glm` by using the first or the last element of the cell array returned by `pspm_load_data` .

Hi @dominikbach Many thanks for the suggestions. I do agree the previous version has not considered selecting marker channels. I have now added an additional option for `glm` for selecting the first/last/specific channel from the loaded data, and defaultly it will select the first `marker` channel returned from `load_data`. Please feel free to let me know if there are further issues that I have not considered. Thanks

dominikbach (approved):

dominikbach (changes requested):
I think in line 977 marker_chan_num is still set to 1 by default, rather than to "marker"?

PR #461, opened 2023-02-07
Fix issue 457 -- sf_dcm

### LINKED ISSUE #457: Enable data with `NaN` for `pspm_sf_dcm`
### Feature Description
Users have found `pspm_sf_dcm` reporting errors when running with data that includes `NaN`s. Two new `options` fields for `pspm_sf_dcm`, `missing` and `missing_threshold`, are expected for allowing this feature.

teddychao:
Fixes #457.

- `pspm_sf`
  - Now allow input with NaN that indicates missing epochs.
  - Missing epochs can be specified with `options.missing` or the automatic NaN detection.
  - Missing epochs will be verified with `options.missingthresh`. When the missing epochs are longer than 2s, the result `out` will be converted to `[]`.


src/pspm_sf_dcm.m
```diff
@@ -128,43 +118,64 @@
...
+ymissing_start = find(diff(ymissing)==1);
+ymissing_end = find(diff(ymissing)==-1);
+if length(ymissing_start) > length(ymissing_end)
+  ymissing_end = [ymissing_end, length(ymissing_end)];
+elseif length(ymissing_start) < length(ymissing_end)
+  ymissing_start = [1, ymissing_start];
+end
+miss_epoch = [ymissing_start(:),ymissing_end(:)];
+flag_missing_too_long = 0;
+if diff(miss_epoch, 1, 2)/sr > options.missingthresh
```
dominikbach, line 145: LHS is a vector - should this be `any()`?

src/pspm_sf_dcm.m
```diff
@@ -128,43 +118,64 @@
...
+  out = [];
+  flag_missing_too_long = 1;
+  warning(...
+    ['Imported data includes too long miss epoches (over ',...
+    num2str(options.missingthresh), 's), thus estimation has been skipped.'])
+end
+options.isYout = ymissing(:)';
+%% 5 Extract parameters
+if ~flag_missing_too_long
```
dominikbach, line 154: if flag_missing_too_long, the output should be determined as well, in particular fields .t .a, .n. .f, .ma which might be used downstream should be set to NaN. It might be useful to include an additional field .message with the warning message for debugging.

src/pspm_options.m
```diff
@@ -385,6 +385,8 @@
     options = autofill(options,'dispwin',                 1,          0                 );
     options = autofill(options,'dispsmallwin',            0,          1                 );
     options = autofill(options,'fresp',                   0.5,        '>', 0            );
+    %options = autofill(options,'missing',                 [],         '*Num'            );
```
dominikbach, line 388: why is this needed?

src/pspm_sf_dcm.m
```diff
@@ -29,6 +29,8 @@
 %    .dispwin:  display progress window (default 1)
 %  .dispsmallwin:
 %               display intermediate windows (default 0);
+%    .missing:  index of missing values to ignore
```
dominikbach, line 32: For `pspm_dcm`, `missing` is a field of `model` correct? Should we do the same here?

dominikbach (changes requested):

src/pspm_options.m
teddychao: Sorry my fault. This was used previously but no longer taking effects. I have deleted this line now.

src/pspm_sf_dcm.m
teddychao: Indeed, `missing` is a field of `model` in `dcm`. The question I had was `sf_dcm` does not have a struct `model`. May I know if I should manually add a struct `model` in `sf_dcm`?

src/pspm_sf_dcm.m
teddychao: In Matlab, a vector is accepted as the condition for if loop, however it is indeed more explicit to use `any()` here, therefore I have added it. Thanks for reminding me of the point. 🙂

src/pspm_sf_dcm.m
teddychao: Sure. I have now allowed `output` if the missing epochs are too long. I have also added the ID for the error message, apologies for that as I should always set the ID for every error message.

dominikbach: It seems the estimation is now taking place regardless of the missing interval duration? This is not what I meant - I meant setting the aforementioned fields to NaN, and storing the warning message inside of the output, so it can be read from the output file.

dominikbach (changes requested):
see threads

src/pspm_sf_dcm.m
teddychao: > It seems the estimation is now taking place regardless of the missing interval duration? This is not what I meant - I meant setting the aforementioned fields to NaN, and storing the warning message inside of the output, so it can be read from the output file.

I see, thanks for clarifying this. I have made a new attempt, hope this version is ok 🙂

src/pspm_sf_dcm.m
teddychao: > OK, this makes sense - no need to create a "model" variable here. Just wondering whether we should lave such data-fields out of "options". I have no strong opinion, it could be a field in options or an additional input argument. It would also have to be forwarded from pspm_sf, which is something to think about.

Thank you @dominikbach . Currently `.missing` as a field in `options` will be relatively more compatible since we need to adjust the formula if we would like to make `.missing` as an independent parameter. Moreover, `.missing` is not a mandatory field here thus it may be a little difficult to put `.missing` as an independent parameter. Since `options` will be transferred when `sf_dcm` is called, `.missing` will be therefore transferred at the same time.

dominikbach (changes requested):
I think lines 151-160 should only be run if ~flag_missing_too_long (ie. you could put everything into the same if-loop).
Running the model (line 153-154) when it is not actually required just increases run time?

dominikbach (approved):

PR #471, opened 2023-03-13
Fix issue 470

### LINKED ISSUE #470: Legacy use of `chan`
### Summary
`pspm_load_data` is still using `chan` to describe channels, which is inconsistent to the remaining functions. Now `chan` is a legacy usage and should be replaced by `channel` across all the functions.

teddychao:
Fixes #470.

- Update the variable name, from `chan` to `channel`.
  - compute_visual_angle
  - convert_ecg2hb
  - convert_hb2hp
  - data_editor
  - data_editor_App
  - display
  - display_App
  - down
  - find_channel
  - find_sounds
  - find_valid_fixations
  - get_biotrace
  - get_physlog
  - get_rf
  - load_data
  - remove_epochs
  - resp_pp
  - scr_pp
- Change file encoding from CRLF to LF, to keep them consistent to the remaining files.
  - files in pspm_cfg.


dominikbach (approved):
The diff view makes it appear that the entire code for all these functions has changed, which is evidently not the case. So it's difficult to judge what exactly has changed. I will trust you on this and approve.

teddychao:
> The diff view makes it appear that the entire code for all these functions has changed, which is evidently not the case. So it's difficult to judge what exactly has changed. I will trust you on this and approve.

Hi @dominikbach Thanks for approving it. I only changed the variable name from "chan" to "channel" since we have all preferred to use the full name now to prevent ambiguity. I also changed the encoding for some files, but I had made no code modification.

PR #482, opened 2023-05-01
Fix issue 551 -- Add NaN filtering for `pupil_pp`

teddychao:
### Summary 
Add `pupil_pp`'s missing value filtering.
This pull request added a `NaN` checking step in `pupil_pp`. The cut off value is set to be 10% (0.1) in `pspm_options`. If the missing values are more than the cut-off value of the full data, then `pupil_pp` will stop and report the error message saying "... has more than .... % missing values."


src/pspm_pupil_pp.m
```diff
@@ -147,19 +151,36 @@
...
+      'options.channel and options.channel_combine data have different units.');
     return;
   end
   if numel(data.data) ~= numel(data_combine.data)
-    warning('ID:invalid_input', 'options.channel and options.channel_combine data have different lengths');
+    warning('ID:invalid_input', ...
+      'options.channel and options.channel_combine data have different lengths.');
+    return;
+  end
+  % Check if one channel is dominated by NaNs
```
dominikbach, line 173: I think what the function should do is slightly different:
- if only one channel is provided, then just give a warning but continue processing
- if two channels are provided, then give a warning and continue processing with the better channel only, do not combine 
The problem the user had initially reported was that if one channel is bad, then the other one will be compromised by the combination.

dominikbach (changes requested):

src/pspm_pupil_pp.m
teddychao: I understand, thank you @dominikbach for the input.
Now I have updated the logic.
1. If the original `data` channel is bad (has too many NaNs that are more than the cut-off), give a warning but continue processing. This is not related to the `data_combine` channel. The behaviour will `not` terminate the function.
2. If the `data_combine` channel is bad (has too many NaNs that are more than the cut-off), give a warning *and* do not combine, but continue processing with the original data channel, this is equivalent to "data_combine is not specified". The behaviour will `not` terminate the function.
3. If the `data_combine` channel is not specified, in this case it will be defined as empty (`[]`), the default value. This will continue processing the original data channel. The behaviour will `not` terminate the function.

My understanding is this can provide the expected behaviours. Please forgive me if I am not correct. Thank you.

dominikbach (approved):

teddychao:
fixed some error display bugs, previously it was not displayed properly
image
 fixed a strange expression error, not sure about why.
fixed the invalid `\` at the end of line
image

dominikbach:
@teddychao thanks for reviewing. Please update GUI & documentation as well.

teddychao:
GUI update:
image

image

teddychao:
Hi @dominikbach I have completed these features. However, I have a question. Currently `chan_valid_cutoff` (originally `nan_cutoff`) is used when data combination operation is used. I am not sure if this value should be used if data combination behaviour is not used by users? For instance, users want to just do preprocessing for a channel they focus on, should `chan_valid_cutoff` be used at all?

dominikbach:
@teddychao Thanks, I have added a warning and slightly re-worded GUI and Manual text.

dominikbach (approved):

PR #483, opened 2023-05-15
Fix Issue 484 --- `pspm_sf`

### LINKED ISSUE #484: Missing epoch issue of `pspm_sf`
### Summary
`pspm_sf` stops working when the imported data contains missing values.

### Technical Info
* PsPM version: develop (12 May 2023)
* MATLAB version: 2022b

teddychao:
Fixes #484.

### Introduction
Missing epochs often appear in the imported data. It is expected to let users know the imported data contain missing epochs since they impact the results of processing. Specifically for this issue, the source data contains missing epochs at the beginning. Since some processing requires the initial values for determine its parameters, such missing epochs finally lead to errors. At the same time, the missing epochs cannot be ignored in VBA processing, and therefore there are further issues inside VBA functions. This pull request aims at resolving such initial missing epochs that disallow following analysis.

### Methods
- When there are missing epochs, let users know missing epochs exist and the results will be affected. 
	- Users can adjust the field `options.missingthreshold` to skip trials that contain missing epochs more than the threshold time.
- Initial missing epochs should be filtered out when selecting initial values for determining some parameters.
- Interpolation will be performed before processing VBA analysis.

### Results
- Overall
  - This pull request has been updated in line with the recent pull request #502 .
- `src/pspm_dcm_inv.m`
	- DCM analysis requires the first three values of the imported data for determining priors. Now it uses non-*NaN- values for determine such parameters (Line 556 --- Line 559).
	- `.missing_data` as a field of `model` has been described in the help text.
	- `.missing` is not used thus removed from help text.
	- `.missing_data` does not require a default value because there may be no missing values.

- `src/pspm_options.m`
  - Line 415: update the `options` for `pspm_sf` to allow users adjust the threshold.
	  - The field will be transferred to `sf_dcm` when `sf` calls `sf_dcm`.

- `src/pspm_sf.m`
  - Line 28--33: help text describes `model.missing` is used for specifying the missing epochs.
  - Line 183--192: `model.missing` is used to load missing epochs.
  - Line 219--238: Loading missing epochs.
  - Line 288--289: use `pspm_time2index` to load missing data, from time ('seconds') into timepoints.

- `src/pspm_sf_dcm.m`
	- Line 115 --- Line 116
		- Now it uses only non-*NaN- values for determining `X0` and `nresp`.
  - Line 154 --- Line 163
    - Now reminds users when there are missing epochs if it is the situation since the results may be dominated with NaN.
    - Now allows users to change the threshold for skipping trials that contain missing epochs if they wish.
  - Line 169
    - Now it uses `pspm_interpolate` to fill the NaNs before using VBA functions.


src/pspm_sf_dcm.m
```diff
@@ -142,11 +143,17 @@
 end
 miss_epoch = [ymissing_start(:),ymissing_end(:)];
 flag_missing_too_long = 0;
-if any(diff(miss_epoch, 1, 2)/sr > options.missingthresh)
-  warning_message = ['Imported data includes too long miss epoches (over ',...
-    num2str(options.missingthresh), 's), thus estimation has been skipped.'];
+if any(diff(miss_epoch, 1, 2)/sr > 0)
```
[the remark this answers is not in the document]
teddychao: Hi @uzaygokay thanks for the comment. The `else` situation of the `if any(diff(miss_epoch, 1, 2)/sr > options.missingthresh)` has another warning message, please check it below...

dominikbach (approved):

src/pspm_sf_dcm.m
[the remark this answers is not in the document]
uzaygokay: Oh sorry thanks for the clarification :)

uzaygokay (approved):

teddychao:
Hi @dominikbach @uzaygokay many thanks for approving this. I will send this updated package to Carlos for him to test. I will merge it when he feels comfortable with the updated code. Thank you

src/pspm_sf_dcm.m
```diff
@@ -105,12 +105,13 @@
...
 % 4.4 determine initial conditions
-x0 = y(1:3);
+y_non_nan = y(~isnan(y));
+x0 = y_non_nan(1:3);
 X0(1, 1)   = mean(x0);
 X0(2, 1)   = mean(diff(x0));
 X0(3, 1)   = diff(diff(x0));
 priors.muX0 = X0;
-nresp = floor(fresp * numel(y)/sr) + 1;
+nresp = floor(fresp * numel(y_non_nan)/sr) + 1;
```
dominikbach, line 114: This line should be reverted. the numbe of modelled responses should not depend on the number of valid data points, only on the duration of the epoch.

src/pspm_sf_dcm.m
```diff
@@ -131,6 +132,26 @@
   '\n=========================================================\n'], c(4:6));
 if isfield(options, 'missing')
   ymissing = options.missing;
+  if isfile(ymissing)
+    [~, ~, fExt] = fileparts(ymissing);
+    switch lower(fExt)
+      case '.mat'
+        % A MAT file
+        ymissing = load(ymissing);
```
dominikbach, line 140: Here, the generic function 'pspm_get_timing' should be used. 

Compare the corresponding code in pspm_dcm: pspm_get_timing('epochs', ...
      model.missing{iSn}, 'seconds');

src/pspm_sf_dcm.m
```diff
@@ -131,6 +132,26 @@
...
+        ymissing = ymissing.epochs;
+      otherwise  % Under all circumstances SWITCH gets an OTHERWISE!
+        warning('ID:invalid_input', 'Missing epochs should be a matlab file if loading from external.');
+    end
+  else
+    if ~isnumeric(ymissing)
+      warning('ID:invalid_input', 'Missing epochs should be numeric if as a variable.');
+    end
+  end
+  if any(size(ymissing)~=size(y))
```
dominikbach, line 150: I'm not sure what this does.

src/pspm_sf_dcm.m
```diff
@@ -131,6 +132,26 @@
...
+    ymiss = ymissing;
+    ymissing = zeros(size(y));
+    ymissing(ymiss(:,1)+1:ymiss(:,2)+1) = 1;
+  end
 else
   ymissing = isnan(y);
```
dominikbach, line 156: I'm not sure this works. Above, 'ymissing' is defined by onset and offset, here it is an index?

Below it seems to be treated not as an index, but as onset and offset definition (e.g. when you use 'diff' to determine the duration of missing epochs).

dominikbach (changes requested):

src/pspm_sf_dcm.m
teddychao: Yes, I was designing `ymissing` to be an index.
I checked the usage of `isYout` in `VBA_NLStateSpaceModel` and I find it to be an _index_. That is why I convert it from _onset and offset_ to be an _index_.
To keep the code consistent to previous versions, I can keep `ymissing` to be *onset and offset* and calculate `isYout` (which is *index*) from `ymissing`.
Sorry for making these codes confusing.

src/pspm_sf_dcm.m
teddychao: Understood. I updated this from `y` to `y_non_nan` because of the incorrect `x0` calculated at line 109 when the initial values are NaN. I do not have knowledge about whether `nresp` should use `y` or `y_non_nan`. I have reverted this line now.

dominikbach:
I realise now there is a more fundamental problem. 

pspm_sf takes one or multiple data files as data inputs, and an epoch definition file as event input. It will then pass the data to pspm_sf_dcm *for each identified epoch*. 

This means that the missing epoch definition needs to be retrieved in pspm_sf (if method is "dcm"), and a missing index needs to be passed on to pspm_sf_dcm.

teddychao:
> I realise now there is a more fundamental problem.
> 
> pspm_sf takes one or multiple data files as data inputs, and an epoch definition file as event input. It will then pass the data to pspm_sf_dcm _for each identified epoch_.
> 
> This means that the missing epoch definition needs to be retrieved in pspm_sf (if method is "dcm"), and a missing index needs to be passed on to pspm_sf_dcm.

This is indeed different from the design in `pspm_dcm` and `pspm_dcm_inv`... Can we open another pull request for this problem after agreeing the changes in this pull request that introduces SF's usage of missing epochs?

dominikbach (changes requested):
Some questions and comments.

src/pspm_sf.m
```diff
@@ -227,39 +263,48 @@
...
       sf.model{k}(iEpoch).timeunits  = model.timeunits;
       sf.model{k}(iEpoch).samples    = win;
       sf.model{k}(iEpoch).sr         = sr(datatype(k));
       %
       escr = Y{datatype(k)}(win(1):win(end));
       sf.model{k}(iEpoch).data = escr;
-      % 3.7 do the analysis and collect results
-      invrs = fhandle{k}(escr, sr(datatype(k)), options);
+      if any(missing{iSn})
+        model.missing_data = zeros(size(escr));
```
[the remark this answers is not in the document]
teddychao: Yes this is my fault.. I have corrected this with `pspm_time2index` and added the update notes. Thanks for the guidance.

src/pspm_cfg/pspm_cfg_sf.m
```diff
@@ -191,7 +193,7 @@
 mrk_chan.name    = 'Marker Channel';
 mrk_chan.tag     = 'mrk_chan';
 mrk_chan.strtype = 'i';
-mrk_chan.val     = {0};
+mrk_chan.val     = {1};
```
dominikbach, line 196: This change is unclear to me.

dominikbach (approved):

PR #485, opened 2023-05-22
Fix issue overwrite

teddychao:
Fixes the overwrite issue that was proposed by users.

### Introduction
The behaviour of overwriting is controlled by two factors: (1) the default values set in `pspm_options`; (2) the checking in `pspm_overwrite`. However, some functions that call `pspm_overwrite` are not overwriting files properly, because the returned values from `pspm_overwrite` were not only the overwriting marker but also the status marker that is used as a general rule in PsPM. This pull request is therefore fixing the behaviour of `pspm_overwrite`.

### Method
The expected behaviour of `pspm_overwrite` is to always overwrite if no corresponding file exists in the path.

### Results
The return value of `pspm_overwrite` is called and used for determine whether to overwrite files. This is corrected for 
* `con2`: line 76---77
* `data_editor`: line 202---203
* `get_marker_info`: line 80---81
* `import`: line 308---309
* `segment_mean`: line 157---158
Also, the default value of overwrite for `pspm_import` has been updated to `1` from `0`.

### DIFF
--- a/src/pspm_con2.m
+++ b/src/pspm_con2.m
@@ -73,7 +73,8 @@ if exist('options','var')
 end
 options = pspm_options(options, 'con2');
 %% check outfile
-if pspm_overwrite(outfile, options) == 0
+ow = pspm_overwrite(outfile, options)
+if ow == 0
   return
 end
 %% assemble input (for diagnostic checking)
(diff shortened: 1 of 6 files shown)


uzaygokay (approved):
Even though I still lack theoretical understanding of the software, updated parts are very straightforward. You basically store the  `pspm_overwrite` function in a variable called ow instead of just calling it and the default value for `pspm_import` has changed.

PR #492, opened 2023-06-19
Fix issue 491 --- missing allocation

### LINKED ISSUE #491: Generalise missing epochs allocation
### Feature Description
Missing epochs can be defined by a field `.missing` in PsPM. It is more appropriate as a field of `model` rather than `options`.

teddychao:
### Introduction
As described in #491, it is currently agreed that the field `.missing` should be put under the struct `model` over `options`. This pull request aims at moving such `.missing` fields to `model` for all relevant functions.
image

### Results
The details of all the functions that use `.missing` or `.missing_data` are listed below:

* `dcm`
  * `.missing` is allocated under `model` as an optional field.
	  * Help text: Line 33
	  * Checking: Line 292
	  * Usage: Line 341
* `dcm_inv`
  * `.missing_data` is used to load missing epoch data that was loaded from `dcm`, as an optional field.
    * Help text: Line 37
    * Checking: Line 131
    * Usage: Line 548
* `find_valid_fixations`
  * Not modified.
    * `.missing` is allocated under `options` because `model` is unavailable under this function.
* `glm`
  * Not modified.
    * `.missing` is allocated under `model` as an optional field.
* `interp1`
  * The function, which originally was written for `prep_data`, performs interpolation for data based on missing epochs and NaNs.
  * The function is now used by `sf_dcm` for interpolating data.
  * The logic of interpolating is
    * The head and tail of data are individually interpolated, if there are NaNs.
    * It will both detects NaNs in data and considers the missing epochs. They could be overlapping but could be different as well. Therefore, the NaNs in data and missing epochs are merged into an index that is used for interpolating.
* `sf`
  * `.missing` is allocated under `model` as an optional field.
  * Checking for `.missing` has been added for `sf`, which can be seen at _Line_ 183 -- _Line_ 192 and _Line_ 218 --- _Line_ 236.
  * Calculating `missing` from `model.missing` can be updated to include "checking for overlap and merging" (code line 223--232). This feature request is currently described in #499 .
  * The calculation of `.missing_data` from `.missing` is assigned at pull request #484.
* `sf_dcm`
  * `sf_dcm` is now using the struct `model`.
  * `.missing_data` is allocated under `model`.
  * Checking for `model` fields is following the style used in `dcm`.
* `split_sessions`
  * Not modified.
    * `.missing` is allocated under `options` because `model` is unavailable under this function.


dominikbach (changes requested):
dcm_inv should be using the missing field.

src/pspm_dcm_inv.m
```diff
@@ -64,7 +64,6 @@
 %   │                 [numeric, default: 0.1, unit: second]
 %   │                 minimum dispersion (standard deviation) for flexible
 %   │                 responses.
-%   ├──────.missing:  data points to be disregarded by inversion.
```
dominikbach, line 67: Please revert, change to "missing_data", and list under "model".

src/pspm_sf.m
```diff
@@ -25,6 +25,7 @@
 %   │                   as 'dcm'
 %   │                   [cell_array] a cell array of methods mentioned above.
 %   ├─────────.filter:  filter settings; modality specific default
+%   ├────────.missing:  index of missing values to ignore
```
dominikbach, line 28: this is not an index, it should be an epoch definition file.

dominikbach (changes requested):
I'd suggest we slightly refactor the call to pspm_sf_dcm to be consistent with other models:

scr -> model.scr
sr -> model.sr
options.missing -> model.missing

pspm_sf_dcm(model, options)

src/pspm_sf.m
teddychao: I will change it and use `get_timing` to load..
Just to confirm. I updated this line based on `pspm_sf_dcm` at line 32. Shall I make `.missing` to always refer to a missing epoch to all the functions?

src/pspm_dcm_inv.m
teddychao: `.missing_data` is now added at line 37 and used in line 548.

src/pspm_dcm_inv.m
```diff
@@ -126,9 +126,9 @@
 try model.meanSCR; catch, model.meanSCR = 0; end
 % These parameters were set with default fallback values but will be
 % determined later by processing (same to pspm_dcm)
-%try model.fixevents; catch, warning('model.fixevents not defined.'); end
-%try model.flexevents; catch, warning('model.flexevents not defined.'); end
-%try model.missing_data; catch, warning('model.missing_data not defined.'); end
+try model.fixevents; catch, warning('model.fixevents not defined.'); end
+try model.flexevents; catch, warning('model.flexevents not defined.'); end
```
teddychao, line 130: @dominikbach I do not quite remember why these checking codes are commented. I think they are used as
`.fixevents`: line 341, 403.
`.flexevents`: line 336, 337, 338, 394, 395, 399.
Therefore I have re-enabled them.

src/pspm_sf.m
teddychao: Sure. In this pull request the objective is to allocate .missing or .missing_data, thus I have enabled them.
For loading and calculating missing_data, I put them in pull request 484, where SF is following the design in DCM.

src/pspm_sf.m
```diff
@@ -199,7 +215,26 @@
...
+      [~, sortindx] = sort(missing{iFile}(:, 1));
+      missing{iFile} = missing{iFile}(sortindx,:);
+      % check for overlap and merge
+      for k = 2:size(missing{iSn}, 1)
+        if missing{iFile}(k, 1) <= missing{iFile}(k - 1, 2)
+          missing{iFile}(k, 1) =  missing{iFile}(k - 1, 1);
+          missing{iFile}(k - 1, :) = [];
+        end
+      end
+    end
```
dominikbach, line 233: In the future, this code (and the same for pspm_dcm and pspm_data_editor) could be included in pspm_get_timing (optional). Could you open a code improvement issue for this when you merge here?

src/pspm_sf_dcm.m
```diff
@@ -18,19 +18,23 @@
...
-%               display intermediate windows (default 0);
-%    .missing:  index of missing values to ignore
-%   .missingthresh:
+%   ┌──────model
+%   │ ▶︎ Mandatory
+%   ├───────.scr:  skin conductance epoch (maximum size depends on computing
+%   │              power, a sensible size is 60 s at 10 Hz)
+%   ├────────.sr:  sampling rate in Hz
+%   │ ▶︎ Optional
+%   └.missing_data: the datafile of missing epochs
```
dominikbach, line 27: this should be changed - it is an index?

src/pspm_sf_dcm.m
```diff
@@ -18,19 +18,23 @@
...
+%
+%   ┌────options:  options structure
+%   ├─.threshold:  threshold for SN detection (default 0.1 mcS)
+%   ├─────.theta:  a (1 x 5) vector of theta values for f_SF
+%   │              (default: read from pspm_sf_theta)
+%   ├─────.fresp:  frequency of responses to model (default 0.5 Hz)
+%   ├───.dispwin:  display progress window (default 1)
+%   ├.dispsmallwin:
+%   │              display intermediate windows (default 0);
+%   └─.missingthresh:
```
dominikbach, line 37: this option must be defined via  pspm_sf but it is not included there?

src/pspm_sf_dcm.m
```diff
@@ -47,19 +51,25 @@
 end
 sts = -1;
 tstart = tic;
+
 %% 2 Check input arguments
-if nargin < 2 || ~isnumeric(sr) || numel(sr) > 1
+% 2.1 set model ---
+try model.scr; catch, warning('Input data is not defined.'); return; end
+try model.sr; catch, warning('Sample rate is not defined.'); return; end
+try model.missing_data; catch, warning('Missing data file is not defined.'); end
```
dominikbach, line 59: "Missing data index" instead of "file"?

src/pspm_sf_dcm.m
```diff
@@ -143,7 +153,7 @@
 end
 miss_epoch = [ymissing_start(:),ymissing_end(:)];
 flag_missing_too_long = 0;
-if any(diff(miss_epoch, 1, 2)/sr > options.missingthresh)
+if any(diff(miss_epoch, 1, 2)/model.sr > options.missingthresh)
```
dominikbach, line 156: if there are nan in the data, they need to be interpolated now, to avoid errors in VBA?

dominikbach (changes requested):
Conceptually approved, but a few suspected code errors & suggested improvements flagged.

src/pspm_sf.m
teddychao: I agree. I have referred this comment and code in the new issue (improvement).

src/pspm_sf_dcm.m
teddychao: This is my fault. This should be the index of missing values / epochs. I have now corrected this line and checked other relevant comments.

src/pspm_sf_dcm.m
teddychao: `.missingthresh` is now used in `pspm_sf_dcm` at line 158 to check if the missing epochs are too long... Currently, `model.missing` (missing epoch file) is calculated into `model.missing_data` (missing epoch index, n x 1) by `pspm_sf` and the latter is transferred to `pspm_sf_dcm`. `model.missing_data` is then calculated into `missing_epoch` (missing epoch start and end value matrix, m x 2, where m is the number of start / end values of missing epochs) by `pspm_sf_dcm`. `.missingthresh` is then used to evaluate the `missing_epoch` (line 158 in `pspm_sf_dcm`) to check if the missing epochs are too long.
I think, if necessary, we can add `.missingthresh` to `pspm_sf` and let `pspm_sf_dcm` to inherit it if that is more intuitive. Currently `pspm_sf` is not using `.missingthresh` because `pspm_sf` does not include the step of converting `.missing_data` into `missing_epoch`.

src/pspm_sf_dcm.m
teddychao: I understood it as, if `.missing_data` is not defined, the only reason is `.missing` (missing epoch data file) in `pspm_sf` isn't loaded. What users may need to do is to load the missing epoch file in `pspm_sf`. It should not be possible if users have loaded missing epoch file but `.missing_data` here isn't added.. However, I agree "missing data index" is more explicit, so I have updated it now.

src/pspm_sf_dcm.m
teddychao: Yes that is right. The idea was not implemented because we previously discussed to use `options.isYout` of `VBA_NLStateSpaceModel` to indicate missing epochs. However one of our recent issues said they were not always considered.. Therefore it is sensible to perform interpolation before using VBA functions.

src/pspm_sf_dcm.m
dominikbach: I understand, but the user is going to look at the help for pspm_sf first, so the option should be mentioned there.

src/pspm_sf_dcm.m
dominikbach: Actually, I just realise, there should be no warning if no missing data exists, because this will be the default situation. IMO the warning can be removed entirely.

src/pspm_sf_dcm.m
dominikbach: The line does not seem to be updated?

dominikbach (changes requested):
some minor questions/comments.

src/pspm_sf_dcm.m
teddychao: It is a little strange, my side shows it is updated, please let me check what happened...
image

src/pspm_sf_dcm.m
teddychao: Sure. The option field `.missingthresh` is now available in `pspm_sf` by `pspm_options` and will be inherited by `pspm_sf_dcm` when it is called.

teddychao:
> some minor questions/comments.

I have also updated the GUI to allow users to add missing epoch files when processing SF model.
However, I think the GUI is used for processing individual files, so the missing epoch file is only allowing one file instead of a vector that includes multiple files.

dominikbach (approved):

Issue #489, opened 2023-06-12
Inconsistency between SF, GLM and DCM


teddychao:
### Summary
There is currently an important inconsistency between `pspm_sf`, `pspm_glm`, and `pspm_dcm`. It is expected that the last channel is selected since it is processed lastly. However, it is unclear they are all behaving as expected as sometimes the first channel is selected. This needs further investigation to keep such consistency.

Issue #490, opened 2023-06-18
GLM select incorrect default channel


teddychao:
### Summary
GLM, SF and DCM should always select the first marker channel and last data channel in default consistently, but they are sometimes selecting the last marker channel and first data channel in the current code.

PR #493, opened 2023-06-19
Fix issue 490 --- generalising first marker channel and last data channel

teddychao:
### Introduction
The pull request attempts to fix issues #489 and #490. There is inconsistency between `dcm`, `glm` and `dcm` for selecting data channels and event channels, which potentially confuses users about how channels are selected. It is agreed that the last data channel and the first marker channel are selected as default. The option `marker_chan_num_event` should no longer be used.

### Methods
- The option `marker_chan_num_event` is removed in `sf`, `glm` and `dcm`.
- The first marker channel / event channel is always selected and no users' customisation is allowed.
- The last data channel / wave channel is always selected.

### Results
* `dcm`
  * Line 359--363: the last data channel is selected
    * `data{iSn}{end}.data` is replaced by `y{iSn}`, and `data{iSn}{1}.header.sr` is replaced by `sr{iSn}`, for avoiding confusion in the following lines.
* `glm`
  * Line 1, 868--875
    * `varagout` is replaced by `glm` where `sts` is omitted. This is because some `try` loops will return if it finds errors, but `varagout` cannot be specified through a direct return unless the code is repeated every time. This explains the updates at Line 868---875.
  * Line 263--286 (old) or Line 264 (new)
    * We have agreed to use the first marker channel, so the `events` is always using the first `data{1}.data{:}`.
  * Line 264 (old)
    * The `data=data{end}` is no longer need because the data has been of length 1, thus it is removed.
* `options`
  * Line 274: `marker_chan_num_event` is no longer needed, thus removed.
* `sf`
  * Line 210--212, 218, 222: the first data channel is replaced by the last data channel
  * Line 269: this has been mentioned in another `sf` pull request, which is `event` should be updated to the corresponding channel.


dominikbach (changes requested):
There is a misunderstanding here: the user report was on wave channels, not marker channels.

src/pspm_sf.m
```diff
@@ -184,14 +185,17 @@
   end
   % 3.3 get and filter data
   [sts_load_data, ~, data] = pspm_load_data(model.datafile{iFile}, model.channel);
-  if sts_load_data < 0, return; end
-  Y{1} = data{1}.data; sr(1) = data{1}.header.sr;
-  model.filter.sr = sr(1);
-  [sts_prepdata, Y{2}, sr(2)] = pspm_prepdata(data{1}.data, model.filter);
+  if sts_load_data < 1, return; end
+  y{1} = data{1}.data;
```
dominikbach, line 189: This is messed up. You use the data from the first channel and the sampling rate from the last channel.

Please change to "data{end}" in line 189, 190. Please replace "data{1}.data" in line 192 with "y{1}".  Please update help text to indicate that *last* data channel is used.

src/pspm_sf.m
```diff
@@ -184,14 +185,17 @@
...
+  sr(1) = data{1}.header.sr;
+  model.filter.sr = data{end}.header.sr;
+  [sts_prepdata, y{2}, sr(2)] = pspm_prepdata(data{1}.data, model.filter);
   if sts_prepdata == -1
     warning('ID:invalid_input', 'Call of pspm_prepdata failed.');
     return;
   end
+  y{iFile} = data{end}.data(:);
```
dominikbach, line 197: The function of lines 197 and 198 is unclear. Remove?

src/pspm_sf.m
```diff
@@ -202,8 +206,8 @@
   % 3.5 Get marker data
   if any(strcmp(model.timeunits, {'marker', 'markers'}))
     if options.marker_chan_num
-      [nsts, ~, ndata] = pspm_load_data(model.datafile, options.marker_chan_num);
-      if nsts == -1
+      [sts, ~, ndata] = pspm_load_data(model.datafile, options.marker_chan_num);
```
dominikbach, line 209: If pspm_sf supports multiple data files then the data file index is missing in line 209.

src/pspm_sf.m
```diff
@@ -212,8 +216,8 @@
           ['Channel %i is no marker channel. ',...
           'The first marker channel in the file is used instead'],...
           options.marker_chan_num);
-        [nsts, ~, ~] = pspm_load_data(model.datafile, 'marker');
-        if nsts == -1
+        [sts_load_data, ~, ~] = pspm_load_data(model.datafile, 'marker');
```
dominikbach, line 219: data file index?

src/pspm_sf.m
```diff
@@ -225,7 +229,26 @@
         return;
       end
     end
-    events = data{1}.data;
+    switch class(options.marker_chan_num_event)
```
dominikbach, line 232: as in other functions, I suggest removing this option entirely and just always use the *first* marker channel.

src/pspm_sf.m
dominikbach: Apparently you changed capital Y into small y. That's fine as far as I can see but please check there is no capital Y left in the function.

dominikbach (changes requested):
Apparently, pspm_dcm also uses the first rather than the last data channel. This should also be changed.

src/pspm_glm.m
```diff
@@ -244,25 +244,26 @@
...
 %% 3 Check & get data
-fprintf('Getting data ...');
 nFile = numel(model.datafile);
 for iFile = 1:nFile
-  [sts, ~, data] = pspm_load_data(model.datafile{iFile}, model.channel);
-  if sts < 1
-    return
-  end
+  % 3.1 User output
+  fprintf('GLM analysis: %s ...', model.datafile{iFile});
```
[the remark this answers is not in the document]
teddychao: I do not quite remember the reason why I added this line.. so I delete this line if it is not needed.

src/pspm_glm.m
```diff
@@ -244,25 +244,26 @@
...
+  % 3.2 Check whether model file exists
```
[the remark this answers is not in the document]
teddychao: > the commented check makes sense but it should be outside the if-loop. Why is it commented?

Because it should be assigned at line 231 together with other option fields.

src/pspm_glm.m
```diff
@@ -244,25 +244,26 @@
...
   fprintf('.');
   if any(strcmp(model.timeunits, {'marker', 'markers','markervalues'}))
-    [sts, ~, data] = pspm_load_data(model.datafile{iFile}, options.marker_chan_num);
-    if sts < 1
+    [sts_load_data, ~, data] = pspm_load_data(model.datafile{iFile}, options.marker_chan_num);
+    if sts_load_data == -1
       warning('ID:invalid_input', 'Could not load the specified markerchannel');
       return
     end
-    if length(data) > 1
```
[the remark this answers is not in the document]
teddychao: I changed this line because previously the following code will use `marker_chan_num_event` to determine which channel to return. If this `data = data{end}` is preserved, it will always select the last channel no matter how `marker_chan_num_event` is specified. However, since now it is determine that `marker_chan_num_event` is no longer needed (https://github.com/bachlab/PsPM/pull/493/files#r1237996411), I can put this line of code back.

src/pspm_glm.m
```diff
@@ -244,25 +244,26 @@
...
-      data = data{end};
-    end
     switch class(options.marker_chan_num_event)
```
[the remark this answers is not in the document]
teddychao: > This does not make sense to me. The option "marker_chan_num_event" can be removed in my opinion.
> 
> Either the user specifies the marker channel number as a numeric (i.e. the n-th channel out of all channels) or the first marker channel is used.

I am not sure these two sentences are meaning the same design.. Do you mean reduce the freedom of `marker_chan_num_event` by making it available as only a numeric? As you said `the user specifies the marker channel number as a numeric (i.e. the n-th channel out of all channels)`?

src/pspm_sf.m
```diff
@@ -9,7 +9,7 @@
 %   ├───────.datafile:  one data filename or cell array of filenames.
 %   ├──────.modelfile:  one data filename or cell array of filenames.
 %   ├─────────.timing:  can be one of the following
-%   │                   - an SPM style onset file with two event types: onset &
+%   │                   - an SPM style onset file with two events types: onset &
```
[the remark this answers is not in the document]
teddychao: This is because I noticed the marker channel is always selected through the first channel (https://github.com/bachlab/PsPM/pull/493/files#r1237991145). In this case it is *a* marker event channel. I understand this as an "event" so I renamed the original `marker_channel_events` to be `marker_channel_event`. However I noticed some test data may expect this to be multiple channels so I revert `event` back to `events`. This sentence is an incident that happens after I made full file replacement where I did not notice this line. I apologise for this mistake.

src/pspm_options.m
```diff
@@ -271,7 +271,7 @@
     options = autofill(options, 'norm',                   0,          1                 );
     options = autofill(options, 'overwrite',              0,          1                 );
     options = autofill(options, 'marker_chan_num',        'marker',   '*Num*Char'       );
-    options = autofill(options, 'marker_chan_num_event',  'first',    '*Num*Char'       );
+    options = autofill(options, 'marker_chan_num_event',  'last',     '*Num*Char'       );
```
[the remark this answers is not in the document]
teddychao: I am a little confused about the suggestion.
I agree that this line can be removed if the user is no longer using `last` or `first` to specify marker channels.
However, you also say "Either the user specifies the marker channel number as a numeric (i.e. the n-th channel out of all channels) or the first marker channel is used." (https://github.com/bachlab/PsPM/pull/493/files#r1237991430). Without specifying `marker_chan_num_event`, I am unsure how I am expected to implement this?
You also suggest to keep "This should be put back in, but changed to data{1}, because we always want to use the first marker channel." at line 264 (https://github.com/bachlab/PsPM/pull/493/files#r1237991145). This seems to mean always the first marker channel should be chosen. I don't quite understand if you would still suggest to allow "user specifies the marker channel number as a numeric", if my understanding of the two sentences are correct?

src/pspm_sf.m
teddychao: > If pspm_sf supports multiple data files then the data file index is missing in line 209.

I don't quite remember when I wrote this but this is definitely incorrect. Yes I have fixed it now. Thanks for pointing out the problem.

src/pspm_sf.m
teddychao: Same to previous

src/pspm_sf.m
teddychao: > Apparently you changed capital Y into small y. That's fine as far as I can see but please check there is no capital Y left in the function.

This is to keep the code consistent with `glm` at line 253 (which is what I mentioned on the PsPM meeting last week, but now I see we were not talking about the same thing).

teddychao: > This is messed up. You use the data from the first channel and the sampling rate from the last channel.
> 
> Please change to "data{end}" in line 189, 190. Please replace "data{1}.data" in line 192 with "y{1}". Please update help text to indicate that _last_ data channel is used.

Sure. I think I just need to confirm whether we should leave freedom for users for this as you said *Either the user specifies the marker channel number as a numeric (i.e. the n-th channel out of all channels)* (https://github.com/bachlab/PsPM/pull/493/files#r1237991430)..

src/pspm_glm.m
```diff
@@ -304,7 +305,7 @@
     isnumeric(model.filter.down) && isnan(model.filter.down)
   model.filter.down = min(sr);
 else
-% 4.2 check value of model.filter.down --
+  % 4.2 check value of model.filter.down --
```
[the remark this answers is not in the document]
teddychao: Yes. I have removed the index in these data since they have been of length 1.

teddychao:
> Apparently, pspm_dcm also uses the first rather than the last data channel. This should also be changed.

Thanks for reminding @dominikbach. I have checked `dcm`, `glm` and `sf` now. They are now consistently using the last data channels / wave channels.

dominikbach (commented):
Generally all issue-related changes look fine to me. However, there is one question on a new behaviour you added here.

dominikbach:
One problem is not addressed here but should be:

In pspm_options, the default value for marker_chan_num should always be 0, not 1.

dominikbach (approved):
I made some further changes as detailed in comments. Please check pspm_options and merge if you are happy.

dominikbach:
Further changes made:
1. pspm_options:
- global handling of value "0" in options.marker_chan_num
- default value for options.marker_chan_num is always 'marker' unless for pspm_merge (where it's converted by the call to pspm_trim)
2. adapted help text in cfg file
3. Removed the case that options.marker_chan_num does not exist in pspm_sf, because this will have already been filled in by pspm_options.

dominikbach:
The legacy behaviour of all functions that use marker channels is:
- marker channel is defined as integer -> use this channel number
-  marker channel is empty or 0 -> use first marker channel

Thus, in the GUI (where the marker channel value is bound to be integer), we can use 0 as default.

In pspm_options, if options.mark_chan_num == 0 then it will be changed to 'marker'. This will make pspm_load_data returning all marker channels, of which the first is then selected by the calling function.

teddychao:
> The legacy behaviour of all functions that use marker channels is:
> 
> * marker channel is defined as integer -> use this channel number
> * marker channel is empty or 0 -> use first marker channel
> 
> Thus, in the GUI (where the marker channel value is bound to be integer), we can use 0 as default.
> 
> In pspm_options, if options.mark_chan_num == 0 then it will be changed to 'marker'. This will make pspm_load_data returning all marker channels, of which the first is then selected by the calling function.

Now I understand this. I might have confused you about the default value for the GUI. I will add this point to the developer’s guide, thank you for the detailed explanation…

PR #496, opened 2023-06-26
Fix issue 495

### LINKED ISSUE #495: Wrong x-tick labels in pspm_rev_dcm
### Summary

The following lines:
xt = get(gca, 'XTick');
set(gca, 'XTickLabel', xt * dcm.input.sr);

should be changed to

xt = get(gca, 'XTick');
set(gca, 'XTickLabel', xt / dcm.input.sr);

such that the units are seconds.

One user reported that the x tick labels can be decreasing - this should be checked with an example data set.

teddychao:
Fixes #495 .

- Update pspm_rev_dcm.m


dominikbach:
Tried this on some models and the plot axis look reasonable.

uzaygokay (approved):
Looks also reasonable to me, here is how it looks on the model that I tried 
feedback

PR #498, opened 2023-07-03
Fix issue 497 -- load_data

teddychao:
### Introduction
Recent patches have resolved the issue of identifying the field `chantype`. However, there are some conflicts in the code of `load_data`. The code of converting `channeltype` into `chantype` should be moved to an earlier place, because there is a loop that assumes `data.header` has a field `chantype` there for checking. In addition, some previous results run with legacy code may have saved the field as `channeltype`. Since then, `load_data` may need to consider some legacy data where there is a field `channeltype`.

### Method
The code of checking `channeltype` is moved to an earlier location.

### Results
Now there is no issue for running data where there is the header field `channeltype`.


src/pspm_load_data.m
```diff
@@ -232,17 +228,23 @@
...
 nflag = zeros(numel(data), 1);
 zflag = zeros(numel(data), 1); % records whether data is empty
 % loop through channels
 for k = 1:numel(data)
-  % check header
+  % 7.2 Check header --
   if ~isfield(data{k}, 'header')
     vflag(k) = 1;
   else
+    % 7.2.1 Convert header channeltype into chantype if there are --
```
teddychao, line 243: The code of converting `channeltype` into `chantype` is moved to this place because this loop assumes that there is a field `chantype` there for checking.

dominikbach (approved):

PR #512, opened 2023-08-11
Fix issue 508

dominikbach:
Fixes # 508.

- changes calls to pspm_ecg2hb to pspm_convert_ecg2hb 
- this function was renamed, but the calls weren't


src/pspm_cfg/pspm_cfg_sf.m
```diff
@@ -193,9 +193,9 @@
 mrk_chan.name    = 'Marker Channel';
 mrk_chan.tag     = 'mrk_chan';
 mrk_chan.strtype = 'i';
-mrk_chan.val     = {1};
+mrk_chan.val     = {0};
```
teddychao, line 196: @dominikbach May I confirm with you about this value please? This value is transferred to the functions as `marker_chan_num`. For marker channel I understand it as the first marker channel is used as default. I also noticed you mentioned the default should be `marker`. I saw your comment in a previous pull request that you are not happy about using `0` here. Since the default value of `marker_chan_num` in `pspm_options` is now `marker`, I would like to listen to your opinion if my understanding is incorrect, thanks.

dominikbach: @teddychao I think this comment belongs to pull request #493; I'll reply there.

teddychao:
@dominikbach I have checked all the `pspm_convert_...` functions. Now there are no CFG files that are still using the old names. I have also corrected some code comments.

teddychao (approved):

PR #517, opened 2023-08-13
Fix issue 516

### LINKED ISSUE #516: Varargout needs to be assigned with default values at initialisation
### Summary
Some PsPM functions use `varargout` to control the output. Inside the function itself, it may terminate when specific conditions are met. In these situations, `varargout` may not have been assigned with initial values and therefore report error at the end. It is recommended to assign default values to `varargout` for every function which uses it at the beginning to avoid reporting errors.

A practical error that appears recently due to this systematic issue is #514 .

teddychao:
Fixes #516 .

## Issue Description
Please see the issue #516 .
This is a theoretical patch.

## Method
The variable `varargout` is defined in every function following the initialisation section.

## Results
- Initialise values for `varargout` in the following functions
  - `dcm` as `[]` in `pspm_dcm`
  - `[]` as `y` in `pspm_filtfilt`
  - `[]` as `channel` in `pspm_find_channel`
  - `[]` as `out_channel` in `pspm_gaze_pp`
  - `[]` as `glm` in `pspm_glm_recon`
  - `[]` as `outfile` in `pspm_merge`
  - `0` as `overwrite_final` in `pspm_overwrite`
  - `struct()` as `pfm` in `pspm_pfm`
  - `data` the input argument as `outdate` in `pspm_prepdata`
  - `[]` as `wavedata` in `pspm_pulse_convert`
  - `[]` as `smooth_signal` and `[]` as `model` in `pspm_pupil_pp`
  - `newfilename` the input argument as the output `newfilename` in `pspm_ren`
  - `[]` as `fighandle` in `pspm_rev_con`
  - `struct()` as `fig` in `pspm_rev_glm`
  - `[]` as `outfield` in `pspm_sf`
  - `[]` as `auc` in `pspm_sf_auc`
  - `[]` as `out` in `pspm_sf_dcm`
  - `[]` as `out` in `pspm_sf_mp`
  - `[]` as `scl` in `pspm_sf_scl`
  - `[]` as `scr` in `pspm_transfer_function`
  - `[]` as `newdatafile` in `pspm_trim`
  - The current PsPM version as `v` in `pspm_version`
  - `struct()` as `outinfos` in `pspm_write_channel`


src/pspm_prepdata.m
teddychao, line 1: Hi @dominikbach For `pspm_prepdata`, I noticed the help text indicates it will accept at least two input arguments. From the test function, in `pspm_prepdata_test` line 35, I noticed it was tested with only one input argument, which is the input data. May I confirm with you whether you would like it to accept only one argument? Thanks

teddychao:
> This is fixing a bug introduced by a previous major change in coding style (use of varargout) which should be discussed further.
> 
> What is the benefit of using varargout rather than fixed output arguments?

Hi @dominikbach. We initially discussed this pull request at https://github.com/bachlab/PsPM/pull/439. I am unsure if I explained it clearly to you at that time.
The idea is, by setting `varargout`, we can have the freedom to get both `sts` (the variable indicating whether the script runs successfully) and `dcm` (for example the result from `dcm`, the result variable) or simply `dcm`. I conceptualise it because some functions, for example, `pspm_glm`, does not return `sts`, and have only `glm`. I understood it as `sts` is a marker variable that is useful to check whether the function runs successfully so it is useful to always have it.
However, if `sts` is not required for all the functions, I can revert this pull request and change those functions back.

PR #522, opened 2023-08-18
fix 504

dominikbach:
Fixes # 504.

- changed indexing in pspm_dcm so that when first event in session is at time 0, it won't be excluded.
- also changed wrong help/comment texts in two places

### DIFF
--- a/src/pspm_dcm.m
+++ b/src/pspm_dcm.m
@@ -50,7 +50,7 @@ function varargout = pspm_dcm(model, options)
 %   ├────.filter: Filter settings.
 %   │             Modality specific default.
 %   ├───.channel: Channel number.
-%   │             Default: first SCR channel
+%   │             Default: last SCR channel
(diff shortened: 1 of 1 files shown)


src/pspm_dcm.m
```diff
@@ -556,8 +553,8 @@
 
   % assign trials to subsessions
   trls = num2cell([sn_trlstart{iSn}, sn_trlstop{iSn}],2);
-  subs = cellfun(@(x) find(x(1) > subsessions(:,2) & ...
-    x(2) < (subsessions(:,3)) ...
+  subs = cellfun(@(x) find(x(1) >= subsessions(:,2) & ...
+    x(2) <= (subsessions(:,3)) ...
```
teddychao, line 557: Sorry I was not aware of this situation when I check the code..

teddychao (approved):

PR #528, opened 2023-09-07
update overwrite logic

teddychao:
This pull request is to update the logic of `pspm_overwrite`.

## Summary
The logic of `pspm_overwrite` is now clearer based on the following aspects:
1. "Overwriting" should only happen when there has been a file with the same name of the file PsPM would like to write. Therefore, if there is no file with the same name, PsPM should always be allowed to write. This is now described at Line 83--85 in `pspm_overwrite`.
2. If PsPM is in develop mode, "overwriting" should always be allowed because the files that may be overwritten are artificial data. This is now set at line 79--82 in `pspm_overwrite`.

### DIFF
--- a/src/pspm_overwrite.m
+++ b/src/pspm_overwrite.m
@@ -73,9 +73,7 @@ switch numel(varargin)
         warning('ID:invalid_input', ...
           'the second input argument should be either a double or a struct.');
         return
-      end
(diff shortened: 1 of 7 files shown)


dominikbach (approved):
Looks good to me.

PR #529, opened 2023-09-08
Fix issue 75

dominikbach:
Fixes # 75.

- possibility to specify a 2-element vector for model.window (mandatory in GUI)
- additional checks on model.window

should be merged after fix-issue-521


teddychao:
I made a few changes which are caused by the previous pull request of overwriting. In general it is, in the develop mode, for example in the GitHub test environment, it is always allowed to overwrite, so there is no longer need to specify `options.overwrite` there. The changes are all included in only one commit to make it easy to revert in case you are not happy with them, and they are also listed below.

teddychao (approved):

PR #541, opened 2023-09-29
fix help text

### LINKED ISSUE #537: pspm_split_sessions behaves unexpectedly when options.suffix == 0
### Summary

When options.suffix == 0, then in line 189 ff the mean marker distance is added as suffix. However, the user might expect that actually no suffix is added.

Having no suffix might indeed be problematic as there is a marker in the last sample of the file. However, this needs to be explained in the help text.

dominikbach:
Fixes #537  .

- updates the help text in pspm_split_sessions


teddychao (approved):

teddychao:
This is similarly a help text only update so I merge it directly.

PR #543, opened 2023-09-30
Fix issue 536

### LINKED ISSUE #536: pspm_scr_pp adds channel with missing_epochs_filename argument
### Summary

In pspm_scr_pp, when options.missing_epochs_filename is provided, then the original data should not be changed:
%   ├.missing_epochs_filename:
%   │             If provided will create a .mat file saving the epochs.
%   │             The path can be specified, but if not the file will be saved
%   │             in the current folder. If saving to the missing epochs file,
%   │             no data in the original datafile will be changed. For
%   │             instance, abc will create abc.mat

Instead, a new channel with the same data is created. The missing epochs file is correctly saved as well.

### Technical Info
* PsPM version: develop branch as of today
* MATLAB version: 2021b

teddychao:
Fixes #536 .

- `scr_pp`
  - Now the function will not change the original file if it saves epochs to `missing_epoch.mat`
  - The default value of `missing_epoch_filename` is now `''` for better determining whether users would like to save `missing_epoch`.
  - The test function for `scr_pp` has been updated now.


src/pspm_options.m
```diff
@@ -404,7 +403,7 @@
         return
       end
     end
-    if options.change_data == 0 && ~isfield(options, 'missing_epochs_filename')
+    if options.change_data == 0 && ~isempty(options.missing_epochs_filename)
```
dominikbach, line 406: Should this be isempty rather than ~isempty?

dominikbach (changes requested):

src/pspm_options.m
teddychao: Apologies, this is my fault here. I have resolved it now by correcting this line of code and adding the corresponding test functions.

dominikbach (approved):

PR #551, opened 2023-10-23
Fix issue 548 -- scr_pp's clipping detection

### LINKED ISSUE #548: scr_pp does not detect artefacts in SCR channel
### Summary
This automatic preprocessing function works perfectly for most of the data files, except for one file with an artefact period. 
### Steps to Reproduce
options = struct();
options.max = 100;
options.missing_epochs_filename = "output filename";
[sts, out] = pspm_scr_pp("data file", options);

### Expected Results
get an output file with variable "epochs" which contains the start and end time of the artefact
### Actual Results
the output epoch is empty
it looks like that the default "absolute SCR values in microSiemens" and "slope per second" criteria cannot detect artefacts with a sudden increase (like, 90 degree increase in data) and decrease. see the example data (attached).

### Technical Info
* PsPM version: 6.0.0
* MATLAB version: 9.13.0.2049777 (R2022b)
[tpspm_REW2_acq_0005_sn02.zip](https://github.com/bachlab/PsPM/files/13044107/tpspm_REW2_acq_0005_sn02.zip)

teddychao:
Fixes #548 .

Users reported `scr_pp` can detect clipping (https://en.wikipedia.org/wiki/Clipping_(signal_processing)) but cannot detect baseline alteration. The proposed solution is to add an option to include baseline alteration into clippings.

- `pspm_scr_pp`
  - New feature: now `scr_pp` can detect baseline filtering. This has to be enabled by setting `options.include_baseline=1`
  -  The baseline detection may need adjustment of detection parameters such as `clipping_window_size`

Detection rule for baseline:
* Use moving window.
* For the windowed signal in every loop, check if there are sudden increase *and* reduction.
* If sudden increase and reduction are detected, find the corresponding index and return as the result of this loop.
* Obtain such time stamps and calculate the detected baseline alteration.


dominikbach (approved):
I haven't tested it but it looks alright to me.

PR #564, opened 2023-12-27
adapt pspm_split_sessions

### LINKED ISSUE #563: pspm_split_sessions drops markers at end of file
### Summary

pspm_split_sessions drops any markers in prefix and suffix periods. However, it also drops markers at the end of the file where no splitting occurs.

If marker channel 1 has 30 markers and breakpoints at 10 and 20, then three sessions will result: from start to 10, from 10 to 20 and from 20 to end.

To enable this, pspm_split_sessions trims the file from 1-10, 11-20, and 21-30, and drops markers in the offset periods.

So if marker channel 2 has a marker that occurs AFTER marker 30 of marker channel 1, then these will be erroneously dropped.

### Expected Results
Markers at beginning or end of file should never be dropped, as marker duplication is not an issue here.

- use pspm_trim to trim at one specific marker but with no trimming at the beginning or end of file (using the option "none").

dominikbach:
Fixes #563 .

- avoids markers being dropped in secondary marker channels before first or after last marker of primary channel during pspm_split_sessions
- as a consequence, no trimming is done for the first and last session

### DIFF
--- a/src/pspm_split_sessions.m
+++ b/src/pspm_split_sessions.m
@@ -50,7 +50,7 @@ function varargout = pspm_split_sessions(datafile, markerchannel, options)
 %   REMARK for suffix and prefix:
 %   Markers in the prefix and suffix intervals are ignored. Only markers
 %   between the splitpoints are considered for each session, to avoid
-%   duplication of markers.
+%   duplication of markers. 
 % ● History
 %   Introduced in PsPM 5.1.1
 %   Written in 2021 by Dominik R Bach (Wellcome Trust Centre for Neuroimaging)
@@ -170,7 +170,8 @@ if isempty(splitpoint)
   return;
 else
   % initialise
-  suffix = zeros(1,(numel(splitpoint)+1));
+  preffix = num2cell(zeros(1,(numel(splitpoint)+1)));
+  suffix = num2cell(zeros(1,(numel(splitpoint)+1)));
   for sn = 1:(numel(splitpoint)+1)
     if sn == 1
       trimpoint(sn, :) = [1, max(splitpoint(sn) - 1, 1)];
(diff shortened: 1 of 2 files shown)


test/pspm_split_sessions_test.m
```diff
@@ -168,8 +172,10 @@ function test_splitpoints(this, splitpoints)
       for i = 1:numel(newdatafile)
         if exist(newdatafile{i}, 'file')
           % test suffix and prefix
-          [~, info, ~] = pspm_load_data(newdatafile{i});
-          this.verifyEqual(info.duration, sess_dur(i), 'RelTol', 0.5);
+          if i > 1 && i < numel(newdatafile)
```
teddychao, line 175: I am not sure about why the situation `i==numel(newdatafile)` wants to be excluded?

src/pspm_split_sessions.m
```diff
@@ -180,20 +181,21 @@
...
+        suffix{sn}  = mean(diff(mrk(trimpoint(sn, 1):trimpoint(sn, 2))));
       end
     else
-      suffix(sn) = options.suffix;
+      suffix{sn} = options.suffix;
+    end
+    prefix{sn} = options.prefix;
+
+    if sn == 1
+        prefix{sn} = 'none'; % don't trim start for first session
```
teddychao, line 196: My understanding is `prefix` and `suffix` are numbers, according to the line 173--174. When they are set to be `none`, my understanding is `none` means "do not perform splitting". In this case, I am not sure if it is good practice to set them to be both `0` to keep them consistent variable types? Just a suggestion.

teddychao:
Hi @dominikbach,
I have some minor suggestions to this version, but they do not have to be accepted. I think the current version looks good to me so I have approved it. I am happy to discuss if you would like to kindly consider my suggestions. Thanks.
Best wishes
Teddy

teddychao (approved):

test/pspm_split_sessions_test.m
dominikbach: Both first and last session are special - no trimming at file start and end. However, I changed the code now to explicitly check session duration also for these sessions in latest commit.

src/pspm_split_sessions.m
dominikbach: `prefix` and `suffix` are used as arguments for `pspm_trim`. This function uses none to indicate no trimming. So, if we'd want to change the convention this would have to be done in `pspm_trim`. I'd suggest however not making this change within the context of this pull request.

dominikbach:
@teddychao If the current test finishes without errors then this can be merged.

teddychao:
> @teddychao If the current test finishes without errors then this can be merged.

Thank you @dominikbach .

I added a minor change to the test function (line 166--169). I think the old version assumes that there are always `split_times` available to obtain the first and last values, but the test shows `split_times` could be empty. I also noticed the line 157 has mentioned a condition when `spiltpoints` is empty, so I added this condition to line 166--169 too. I hope this makes sense to you. Other than this issue, the code can run successfully without problems. I can merge it if you are happy with this update. I fear messing up so I would like to wait your confirmation before merging. Thanks.

dominikbach:
LOoks good to me.

PR #566, opened 2023-12-31
fix issue 565

### LINKED ISSUE #565: Matrix for computing mean response in non-linear model is incorrectly padded with zeros
### Summary

In pspm_dcm, trial-by-trial responses are extracted for computing mean response and thereby, response amplitude priors. When trials have unequal length, the average response is (implicitly) padded with zeros. NaN should be used instead.

It would also be preferable to initialise matrix D with NaN as this will speed up the extraction. 

Line 645-660 of pspm_dcm.

dominikbach:
Fixes #565.

- initialise data matrix with NaN before inserting data of variable size.

### DIFF
--- a/src/pspm_dcm.m
+++ b/src/pspm_dcm.m
@@ -644,7 +644,17 @@ end
 
 % 5.4 extract data from all trials
 winsize = floor(model.sr * min([proc_miniti 10]));
-D = []; c = 1;
+% check maximum trial duration
+maxtrial = NaN(numel(model.scr), 1);
+numtrials = NaN(numel(model.scr), 1);
+for isbSn = 1:numel(model.scr)
+    trialduration = model.trlstop{isbSn} - model.trlstart{isbSn};
+    maxtrialduration(isbSn) = max(trialduration(:));
+    numtrials(isbSn) = numel(trialduration);
+end
+% initialise data matrix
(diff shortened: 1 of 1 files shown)


teddychao:
I can understand the logic of the newly added initialisation code. Just have two points. (1) I tried to test `dcm` with the current test function but I don't think the test function can check the newly added code, so I am not sure if you would like me to add tests for them. (2) I checked line 594 also has an initialisation command `D = [];` for `D`. I dont know if the D there should be initialised too, and whether the `D` at line 594 is identical to the `D` at line 656?

teddychao (approved):

dominikbach:
> (1) I tried to test `dcm` with the current test function but I don't think the test function can check the newly added code, so I am not sure if you would like me to add tests for them. 

There were no tests for the previous version either. I'd suggest making the bugfix now for the upcoming release, and thinking about more rigorous testing in the future.

> (2) I checked line 594 also has an initialisation command `D = [];` for `D`. I dont know if the D there should be initialised too, and whether the `D` at line 594 is identical to the `D` at line 656? Other than the two points, I have approved this pull request.

`D` in line 594 is filled with vectors of fixed length (note that `winsize` is determined outside the two `for` loops) and so there should not be an issue here (although the loop would be faster to initialise `D` anyway).

PR #568, opened 2023-12-31
Use missing information for dcm in pspm_extract_segmenets

### LINKED ISSUE #567: pspm_extract_segments should respect missing data information
### Summary

Currently, pspm_extract_segments for GLM and DCM uses the entire input data, rather than excluding missing data as provided in the model structure.

dominikbach:
Fixes #567 .

- fix only for DCM. Function should be refactored entirely in the future.


teddychao (approved):
I agree with the update.

PR #572, opened 2024-01-03
initial commit to fix #571

### LINKED ISSUE #571: pspm_dcm does not use missing data periods shorter than the threshold
### Summary
pspm_dcm drops information on missing data epochs shorter than the threshold.

### Steps to Reproduce
line 249: data is written into a variable `y`: `y{iSn} = data{iSn}{end}.data;`
In line 297, missing epochs are converted to NaN in the data: `y{iSn}(flanks(1):flanks(2)) = NaN;`

After taking care of missing epochs longer than the threshold, missing epochs are identified from the data in line 361-362. However, instead of using the variable `y` previously modified, the data are again read from the original data structure:
```
sbSn_data = data{sbSn(1)}{1}.data(flanks(1):flanks(2));
sbs_miss = isnan(sbSn_data);
```

As a result, non-interpolated artefacts show up in the model inversion.

dominikbach:
Fixes #571  .

- fix a bug in pspm_dcm that leads dropping sub-threshold missing data periods
- enable combining missing data epochs from file and data (as in a recent use case in CogLearn)

### DIFF
--- a/src/pspm_dcm.m
+++ b/src/pspm_dcm.m
@@ -362,7 +362,7 @@ for vs = 1:numel(valid_subsessions)
   isbSn = valid_subsessions(vs);
   sbSn = subsessions(isbSn, :);
   flanks = pspm_time2index(sbSn(2:3), data{sbSn(1)}{1}.header.sr);
-  sbSn_data = y{sbSn(1)}{1}.data(flanks(1):flanks(2));
+  sbSn_data = y{sbSn(1)}(flanks(1):flanks(2));
   sbs_miss = isnan(sbSn_data);
 
   if any(sbs_miss)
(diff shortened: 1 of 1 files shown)


src/pspm_dcm.m
```diff
@@ -358,7 +362,7 @@
   isbSn = valid_subsessions(vs);
   sbSn = subsessions(isbSn, :);
   flanks = pspm_time2index(sbSn(2:3), data{sbSn(1)}{1}.header.sr);
```
teddychao, line 364: Just a suggestion, should `data{sbSn(1)}{1}.header.sr` be `sr{sbSn(1)}`?

teddychao:
My understanding is this pull request would like to make the code between lines 278 and 298 to be working regardless of the condition of `missing{iSn}`. This makes sense to me. I only have a suggestion of line 360, but it is not a strong opinion.

teddychao (approved):

src/pspm_dcm.m
dominikbach: This is not something I have changed; so I'd suggest leaving this for a separate pull request.

PR #575, opened 2024-01-06
Implement model checks for trial-average model

dominikbach:
Implements the model and options checks for the new trial average model.


src/pspm_check_model.m
```diff
@@ -399,11 +384,145 @@
...
     end
-
 end
 
+%% 9. TAM-specific checks
+% -------------------------------------------------------------------------
+if strcmpi(modeltype, 'tam')
+ if (~isfield(model, 'window') || isempty(model.window) || ~isnumeric(model.window))
+      warning('ID:invalid_input', 'Window is expected to be a numeric value.'); return;
+ elseif
```
teddychao, line 394: I am not sure if it means `else` here, or there would like to be a condition following `elseif`?

teddychao:
Hi @dominikbach 
I think I have fixed the error. It seems to be caused by the `elseif` mentioned above. If you are happy with the update, I can approve this pull request. Thanks
Best wishes
Teddy

dominikbach:
Thank you. Yes this was just a typo.

teddychao:
My understanding is this pull request would like to update the old `pfm` to `tam` and add some related updates. Because I was not the author of `pspm_check_model`, I think I should just approve it.
However, there are some updates for the documents in this pull request if `pfm` is now updated. I update the documents now.

teddychao (approved):

Issue #574, opened 2024-01-06
Pupil preprocessing sequence problems


dominikbach:
### Summary

Users have reported problems with certain sequences of pupil preprocessing steps.

https://github.com/bachlab/PsPM/discussions/573

Problem:
1. Input channels for `pspm_pupil_pp` and `pspm_pupil_correct` are checked in `pspm_options`. Here, it is not allowed to specify channels of type `pupil_pp`, or to give a channel number. All PsPM functions should at least allow channel numbers.
2. Using the generic data type pupil does not work for `pspm_pupil_pp` because left and right pupil channel need to be specified individually.

Questions:
1. Do we need `pupil_pp` and `gaze_pp` channel types? Generally, PsPM does not use a specific "pp" channel type for preprocessed or filtered data, with the exception of `emg_pp` which is integrated data that clearly has a different interpretation from the raw data. 
3. Do we need the non-standard backroom function `pspm_load_single_chan`? In any case it should be converted to a regular PsPM function and tested appropriately.

Proposed solution:
1. For the upcoming release v6.1.1, make sure all pupil preprocessing functions create channels of type `pupil` rather than `pupil_pp`.
2. For future release, remove pupil_pp channel type altogether (requires changing `pspm_init`, `pspm_load_data`, `pspm_load_single_chan`.
3. For future releases, refactor `pspm_load_single_chan` into a regular PsPM function.

dominikbach:
This has been solved by pull requests #623, #621, #620.

PR #576, opened 2024-01-08
Ensure pupil processing functions produce pupil, not pupil_pp, channel output

dominikbach:
Addresses the first point in #574 but not the other two.


teddychao:
Let me fix this...

PR #562, opened 2023-12-18
Fix issue 559

teddychao:
### Summary
This pull request aims at
1. Solving the issue #559.
2. Review all the `cfg_run` functions for preparing the next release.
3. Make sure all the `cfg_run` functions initialise `options` before using it and check fields are available in `job` before transferring data to `options`.

### New functions
There is now a new function `pspm_update_struct`. The function can assess the `struct` fields and transfer the information to the new struct when the field is available.

### New test function
There is now a new text function `pspm_update_struct_test`. The function tests whether `pspm_update_struct` can transfer field values correctly.

### Function-specific updates
| Function | Updates |
|:--|:--|
| `pspm_cfg_run_artefact_rm` | Use `pspm_update_struct` (lines 4, 19--24) |
| `pspm_cfg_run_contrast1` | Use `pspm_update_struct` (lines 18) |
| `pspm_cfg_run_contrast2` | Use `pspm_update_struct` (lines 37) |
| `pspm_cfg_run_data_convert ` | Use `pspm_update_struct` (lines 18--19) |
| `pspm_cfg_run_dcm` | Initialise `options` (line 12). Use `pspm_update_struct` (lines 82, 90--102) |
| `pspm_cfg_run_downsample` | Initialise `options` (line 3). Use `pspm_update_struct` (line 9) |
| `pspm_cfg_run_ecg_editor` | Use `pspm_update_struct` (line 20--22) |
| `pspm_cfg_run_export` | Use `pspm_update_struct` (line 15) |
| `pspm_cfg_run_extract_segments` | Use `pspm_update_struct` (line 45) |
| `pspm_cfg_run_find_sounds` | Initialise `options` (line 5). Use `pspm_update_struct` (line 18, 24) |
| `pspm_cfg_run_find_valid_fixations` | Use `pspm_update_struct` (line 23, 33, 46) |
| `pspm_cfg_run_gaze_convert` | Initialise `options` (line 4). Use `pspm_update_struct` (line 5, 8) |
| `pspm_cfg_run_get_markerinfo` | Use `pspm_update_struct` (line 9) |
| `pspm_cfg_run_glm` | Initialise `options` (line 5). Use `pspm_update_struct` (line 72, 107) |
| `pspm_cfg_run_glm_sps` | The `[]` at line 22 and 25 (were 35 and 38) for `soa` was removed because it does not seem to be necessary. |
[...]

PR #584, opened 2024-01-17
Release 6.2

teddychao:
This pull request can be dynamically updated for release 6.2.
Some documents can be updated under this pull request, like replicating the tutorial tests.


teddychao:
This branch should not be used because it has included #562 that has not been approved.

PR #588, opened 2024-01-18
fix 587

### LINKED ISSUE #587: Problem running single-session DCM
### Summary
DCM does not run from GUI.

### Steps to Reproduce
1. Use non-linear tutorial instructions

### Expected Results
Model runs through

### Actual Results
Model fails (see screenshot)
image

`pspm_check_model` wrongly turns a cell array of char into a cell array of cell arrays in line 209.

### Technical Info
* PsPM version: 6.1.1
* MATLAB version: 2023a

dominikbach:
Fixes #587  .

- fix `pspm_check_model`
- 
-

### DIFF
--- a/src/pspm_check_model.m
+++ b/src/pspm_check_model.m
@@ -201,7 +201,7 @@ if ~isfield(model, 'timing')
 else
   if ~isempty(model.timing)
     if ~iscell(model.timing) || ...
-      strcmpi(modeltype, 'dcm') && ~iscell(model.timing{1} && ~ischar(model.timing{1}))
+      (strcmpi(modeltype, 'dcm') && ~iscell(model.timing{1}) && ~ischar(model.timing{1}))
       % for DCM, model.timing is either a file name or a cell array of
       % events, or a cell array of file names or cell arrays, so we need to
       % take care of cases where model.timing is a cell array but not a cell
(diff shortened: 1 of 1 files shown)


uzaygokay (approved):

src/pspm_check_model.m
```diff
@@ -201,11 +201,11 @@
 else
   if ~isempty(model.timing)
     if ~iscell(model.timing) || ...
-      strcmpi(modeltype, 'dcm') && ~iscell(model.timing{1})
+      (strcmpi(modeltype, 'dcm') && ~iscell(model.timing{1}) && ~ischar(model.timing{1}))
       % for DCM, model.timing is either a file name or a cell array of
       % events, or a cell array of file names or cell arrays, so we need to
```
teddychao, line 206: I am a little confused about this... It says `model.timing` can be a cell array of filenames, but it cannot be a cell array of char. I think a filename is a char..?

dominikbach: Hi Teddy, the comment does not actually say that "`model.timing` cannot be a cell array of char". Of course it can. The comment only says that we need to find the case where `model.timing` is a cell array, but NOT a cell array of char or of cell arrays (which are both perfectly permissible)

PR #592, opened 2024-01-19
Fix issue 590 -- convert_hb2hp

### LINKED ISSUE #590: HB2HP using wrong heart beat channel
### Summary
In the version 6.1.1. HB2HP uses the first, not the last heart beat channel. This is unexpected and should be fixed.

### Technical Info
* PsPM version: 6.1.1.
* MATLAB version: 2023b.

teddychao:
Fixes #590 .

- pspm_convert_hb2hp
  - Make `hb2hp` to select the last channel
    - Change the logic of heart rate channel selection. Only the last channel should be selected, if there are multiple heart rate channels.

## States
Let me test it with artificial data before making it available to review...

### DIFF
--- a/src/pspm_convert_hb2hp.m
+++ b/src/pspm_convert_hb2hp.m
@@ -67,8 +67,8 @@ if nsts == -1
   return;
 end
 if numel(data) > 1
-  fprintf('There is more than one heart beat channel in the data file. Only the first of these will be analysed.');
-  data = data(1);
+  fprintf('There is more than one heart beat channel in the data file. Only the last of these will be analysed.');
+  data = data(end);
 end
 
 % interpolate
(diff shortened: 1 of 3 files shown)


teddychao:
Hi @dominikbach 
Thanks for approving this change. It is relatively minor and I think it should be quite safe.
I checked the test functions `pspm_convert_hb2hp_test`, and I found it almost doing nothing. It tested some uncommon input variables which are quite unlikely to happen for users. The test files under `ecg2hb`, specifically `test_ecg_outlier_data.mat`, `test_ecg_outlier_data_short.mat` and `tpspm_s102_s1.mat`, do not have hb channels at all. The `test_hb2hp_data1.mat` and `test_hb2hp_data2.mat` do not exist in the test data.
I am wondering how urgent it is. If I can be given some test file, or just artificial, I can update this test suit. If this is urgent and the update is quite simple, I can directly merge it, and then I will open a new issue of the test function.

dominikbach:
I'd suggest merging this and adjusting test functions in a separate issue. Smaller pull requests are easier to manage.

PR #595, opened 2024-01-21
fix bug in selection of marker channel in pspm_glm

dominikbach:
# Proposed changes
- fix a bug in pspm_glm that leads to markers being taken from the first marker channel, but markervalues from the last
- this is urgent and should be included in the upcoming hotfix


teddychao:
Many thanks @dominikbach. I include this update in the hot-fix 6.1.2.

teddychao (approved):

PR #611, opened 2024-01-25
new function to convert missing epochs to a logical index

dominikbach:


teddychao:
The newly added test function needs to be added to the test suit. I can do it here.

src/pspm_epochs2logical.m
```diff
@@ -0,0 +1,32 @@
...
+%   Introduced in PsPM 6.2
+%   Written in 2024 by Dominik Bach (Uni Bonn)
+
+if nargin > 2 
+    epochs = pspm_time2index(epochs, sr);
+end
+
+index = zeros(datalength, 1);
+if ~isempty(epochs)
+    for k = 1:size(epochs, 1)
```
teddychao, line 27: I have a better solution...
Simply:
`C=cellfun(@(x,y)x:y,num2cell(epochs(:,1)),num2cell(epochs(:,2)),'uni',false)` and `index([C{:}])=1`.
I dont quite understand what `index=index(1:datalength)` means...

teddychao (commented):
Think my solution may be faster? Just for your information

src/pspm_epochs2logical.m
dominikbach: This might work but it is a lot more complicated to read - and speed is not an issue here.

teddychao (approved):

PR #612, opened 2024-01-25
new ppg2hb functionality using Heartpy library

### LINKED ISSUE #580: Implement new PPG2HB method robustly

dominikbach:
Fixes #580 .

This implements the new heartpy method for `pspm_convert_ppg2hb`. 

In this or a different pull requests, appropriate tests should be added.


src/pspm_convert_ppg2hb.m
```diff
@@ -12,162 +12,227 @@
...
+%% Heartpy
+if strcmpi(options.method, 'heartpy')
+    % initialise python
+    if isempty(options.python_path)
+        sts = pspm_check_python;
+    else
+        sts = pspm_check_python(options.python_path);
+    end
+    if sts < 1, return; end
+    pspm_check_python_modules('heartpy');
```
teddychao, line 99: Should this function give a return value that explains whether this module can be loaded?

src/pspm_convert_ppg2hb.m
```diff
@@ -12,162 +12,227 @@
...
+    TStep = 1; % moving step: 1s
+    Tolerance = 0.5;
+    HRCInit = 60; % init heart rate: 60bpm
+
+    filtered_ppg = py.heartpy.filter_signal(ppg, ...
+        pyargs('cutoff', [1,20], ...
+        'filtertype',  'bandpass', ...
+        'sample_rate', sr, ...
+        'order', 3));
+    filtered_ppg = double(py.array.array('d',(filtered_ppg)));
```
teddychao, line 116: I had some PPG data which is saved on my UCL one drive folder, but I have no access to it atm, apologies

src/pspm_check_python_modules.m
```diff
@@ -28,4 +29,8 @@
     catch ME
```
teddychao, line 29: Do we want this function to give a print message for users to understand whether the module can be loaded, or do we also want to allow this function to tell PsPM whether python package exists or can be loaded?

teddychao:
image
I fixed some bugs and it seems to be working now. I am not sure if the code was validated with another dataset so it was written in that way. If someone had validated before, please test again to make sure I am not introducing incompatibility..
However, I find line 72 it says `ppg=data.data`, and some code following that still says `data{1}.data`. It seems to be a little confusing to me.

teddychao (changes requested):
I would suggest to request @uzaygokay to test with his data on this version of code.. not sure if my tests are enough

teddychao:
In `pspm_cfg_pp_heart_data`, at line 280, the name of `ppg2hb.name` is defined as "Convert Peripheral pulse oximetry to Heart Beat". I think it means PPG? I am not familiar with "Peripheral pulse oximetry", but since I am no longer in that field, I just hope to confirm with you that this name is the expected name...

src/pspm_convert_ppg2hb.m
```diff
@@ -12,162 +12,227 @@
...
+    if isempty(options.python_path)
+        sts = pspm_check_python;
+    else
+        sts = pspm_check_python(options.python_path);
+    end
+    if sts < 1, return; end
+    pspm_check_python_modules('heartpy');
+    
+    % prepare detection
+    ParametersSTFT = struct('TWindow', 16, ...
```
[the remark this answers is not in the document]
dominikbach: fixed.

src/pspm_convert_ppg2hb.m
dominikbach: fixed.

dominikbach:
1. I will test with CogLearn data next week. Please approve if you are happy with the code; I will add fixes if I encounter any problems.
2. I would replace "peripheral pulse oxymetry" with "pulse oxyymetry", this is shorter and clear enough. 
3. GUI only implements one method, could you implement the selection of classic/heartpy as well (as a separate conversion I would suggest, similar to the choice of Pan-Tompkins/AMRI for ECG2HB).

dominikbach:
I have now tested this on real data and made minor improvements. From my side, this is ready to be reviewed again. 
Please add GUI implementation & tests.

dominikbach:
@teddychao Regarding GUI I think it would be better to merge this, then merge #666 and then add the GUI implementation for heartpy.

teddychao (approved):

PR #628, opened 2024-02-16
new function that checks/creates a python environment.

### LINKED ISSUE #617: Function which check if python environment exists
### Feature Description
write a matlab function that 
(a) checks whether a python environment exists, and 
(b) if not, creates one from the specification of a path to a python installation.

MadniAbdulWahab:
Fixes #617  .

- A new function that
(a) checks whether a python environment exists, and
(b) if not, creates one from the specification of a path to a python installation.


dominikbach:
Two suggestions:
- name the function `pspm_check_python`
- throw a warning if no python environment could be created, and return a status argument `sts` with values `0` or `1`.

PR #632, opened 2024-02-16
Fix a bug in pspm_down

dominikbach:
Fix a bug detected in a use case.


teddychao (approved):

teddychao:
This is a minor update. I merge it directly because there is no change from me. Thanks.

PR #633, opened 2024-02-19
fix DCM bug for missing epcohs

dominikbach:
Fixes a bug in `pspm_dcm` detected in a use case.

After finding missing epochs and splitting data & design in valid and invalid subsessions, only data & design from valid subsessions is selected in lines 536 ff. 

The same selection was made again in line 599 and 649 ff which would not work if there were any invalid subsessions.


teddychao:
Just to confirm `proc_subsessions` is also used in line 595--597. `model.scr` has been `sbs_data(proc_subsessions)` (line 541), and line 597 says it is selected again as `model.scr(proc_subsessions)'`. If line 599 has eliminated `proc_subsessions` for `model.scr`, is line 597 also necessary?
I tested it with test functions but I feel this line is not tested with test function, so we may have to manually check here.. 
Thanks

teddychao (approved):

PR #636, opened 2024-02-22
Enable denoising for all continuous marker channels

### LINKED ISSUE #634: Implement denoising of continuous event channels
### Feature Description

There is a function `pspm_denoise_spike` that filters too short events for specific smr settings. This should be generalised to operate on any continuous event channel, and called from `pspm_get_events`

dominikbach:
Fixes #634.

- enable denosing of event channels other than smr format
- change format of argument `denoise` to specification in seconds, rather than milliseconds, as we want to use SI units wherever possible
- adapt tests


teddychao:
Sorry I spent a few minutes to understand what was going on there... The basic logic is to use `lo2hi` and `hi2lo` to identify pulses, and then use `.denoise` to validate those `pulses`. If `.denoise` is specified, only the `lo2hi` and `hi2lo` that allow longer duration (than `.denoise`) can be accepted for following processing. Hope my understanding is correct.

teddychao (approved):

PR #643, opened 2024-02-27
refactor get-timing & glm

dominikbach:
`pspm_glm` re-sorts conditions across multiple sessions. The same functionality is required for `pspm_extract_segments`. Hence, I moved this functionality to `pspm_get_timing` from where it is accessible to both functions.


teddychao:
# 10 March 2024

#643

* pspm_get_timing
	* Line 227 -- 235
		* The function now has a new variable `allnames` to record the name from all sessions. The logic is sensible to me.
		* I tested how line 228 to line 234 works. I find the variable `allnames` can be updated when there are new values from `in.names{iCond}`. The updated `allnames` becomes an array, such as `{'a1','a2'}`. I expect `in.names{iCond}` to be always a simple string so that the `strcmpi` can be executed to see if `allnames` has included `in.names{iCond}`.
	* Line 298 -- 315
		* It seems to be checking if a condition (the session name is recorded in `allnames{iCond}` and is unique in it) and then obtain the tags from `temptiming(iFile)`. I find the logic makes sense to me.
		* I think line 308 -- 310 seems to be a lower-level if clause under the condition of line 305. If my understand is correct I suggest to indent line 308--310 to a lower level.
* `pspm_glm`
	* Can I confirm the condition of line 398--415? It looks like lines 395--414 are all related to the condition of `iSn`. The current logic will make line 396 be executed at the same condition of line 409--414, but line 396 is separated written. Would it be easier to read if more line 396 to the level of line 409--414?

teddychao (approved):

dominikbach:
Thanks @teddychao for spotting these stylistic points - they are fixed now.

PR #645, opened 2024-02-29
fix DCM for NaN data

### LINKED ISSUE #644: `pspm_dcm` fails if all data are NaN
### Summary
The call ` [post, out] = VBA_NLStateSpaceModel(model.meanSCR(:)',u,f_fname,g_fname,dim,invopt);` in `pspm_dcm_inv`
 returns empty variables; the subsequent extraction fails.

Test data: sub 3 from  https://zenodo.org/records/5573788

### Solution
There should be no valid subsessions in this case. If there no valid subsessions, `pspm_dcm_inv` should not be called in the first place.

dominikbach:
Fixes #644  .

- account for an edge case when all data are NaN


dominikbach:
@teddychao Could you kindly add a test case with only NaN data?

teddychao (approved):

teddychao (changes requested):
The original `isempty` in line 277 is not quite working. At line 275, `isnan` will return an array such as `[0,1,0]`, in which `0` and `1` are logical. However, if we type `isempty(isnan([1,2,3]))`, it will return `0`, which means `isnan([1,2,3])` is not empty, and there is `nan` in `[1,2,3]`. This is not correct. If we use `any(isnan([1,2,3]))`, we also get `0`, but this correct, because there is no `nan` in `[1,2,3]`.

teddychao (approved):

teddychao:
@dominikbach I fixed the issues I mentioned above in the commits. I also added the test for missing values as `test_hra1_flex_cs_missing`. Maybe you can have a look and merge if you are happy with them. Thanks

PR #650, opened 2024-03-05
Updated pspm_leaky_integrator and pspm_pp

MadniAbdulWahab:
Fixes # 647.

- used snake case instead of camel case.
- Converted tau from seconds to samples using the sample rate in pspm_pp

### DIFF
--- a/src/pspm_pp.m
+++ b/src/pspm_pp.m
@@ -108,8 +108,8 @@ switch method
       warning('ID:invalid_input', 'Tau must be numeric.'); return;
     end
     % Convert tau from seconds to samples
-    sampleRate = data.header.sr;  % Assuming the sample rate is stored here
-    tau_samples = tau_sec * sampleRate;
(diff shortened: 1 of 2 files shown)


src/pspm_pp.m
```diff
@@ -96,6 +102,20 @@
...
     if sts == -1, return; end
     msg = sprintf('butterworth filter');
+  case 'leaky_integrator'
+    tau_sec = varargin{4}; 
+    if ~isnumeric(tau_sec)
+      warning('ID:invalid_input', 'Tau must be numeric.'); return;
+    end
+    % Convert tau from seconds to samples
+    sample_rate = data.header.sr;  % Assuming the sample rate is stored here
+    tau_samples = tau_sec * sample_rate;
```
dominikbach, line 112: I think you need to `round` here and make sure the result is not zero. You could simply use `pspm_time2index`.

dominikbach (changes requested):

dominikbach (approved):

PR #651, opened 2024-03-07
fix overwriting in data editor

dominikbach:
Fixes a user-reported bug in overwriting existing files in `pspm_data_editor`.


teddychao (approved):

teddychao:
Apologies this is my fault. I can merge it here since this is a simple error of mine.

PR #658, opened 2024-03-14
Update python handling and tests

### LINKED ISSUE #649: create tests for `pspm_check_python` and `pspm_check_python_modules`

MadniAbdulWahab:
Fixes #649.

### Summary
- Created test functions for pspm_check_python and pspm_check_python_modules.

### New files
- `src/py_find_location.py`
  - This python file can detect the location of python source in the system.
- `pspm_check_python.m`
  - The file can check if python environment has been set up in the system. 
- `test/pspm_check_python`
  - The test function can automatically detect python in any system by using `pspm_find_python`. No user specification is required. However, it needs that Python has been installed in the system.
- `test/pspm_check_python_modules_test`
  - Due to complexity, I disabled the checking of "no module". The Python library in the GitHub environment seems to have got some packages, and it is tricky to remove them first.
- `src/pspm_find_python.m`
  - This is a new file that can detect python package in the system with two modes
    - Mode 1: automatically load from system. This mode only relies on `src/py_find_location.py` and do not require users to specify first.
    - Mode 2: it load python from a specific path. If users have previously defined python path (in this case the path will be saved as py_loc.txt), the function can use this saved text file to load python.
  - This function has been updated to support both windows and unix.

### Minor changes
- `main.yml` (Github actions)
  - Python environment was set up by updating the main.yml file.
- `.gitignore`
  - The Python location file has been ignored.
- `src/pspm_cfg/pspm_cfg_python.m`
  - The description of Python environment has been adjusted. Users do not need to specify Python before they can use the auto detection feature. The python detection code is done with Python's own script and does not need users to specify first.
- `src/pspm_cfg/pspm_cfg_run_pp_heart_data.m`
  - Now the auto mode can detect python correctly.


test/pspm_check_python_modules_test.m
```diff
@@ -0,0 +1,50 @@
+classdef pspm_check_python_modules_test < matlab.unittest.TestCase
+    % unittest class for the pspm_check_python_modules function
+    % PsPM TestEnvironment
+    % ● Authorship
+    % Abdul Wahab Madni 2024(Uni Bonn) 
+
+    methods (TestMethodSetup)
+        function addFunctionPath(testCase)
+            % Add the path to the source directory
+            src_path = fullfile(pwd, '..', 'src');
```
teddychao, line 10: In my test this does not seem to work.. I saw `~/GitHub/bachlab/PsPM/../src`. This path does not seem to be recognised by MATLAB.

test/pspm_check_python_test.m
```diff
@@ -0,0 +1,62 @@
...
+            else
+                this.assumeFail('No Python environment set up for testing reporting current environment.');
+            end
+            pyenv('Version', original_env.Version);  % Reset to original Python environment after the test
+        end
+
+        function test_set_new_python_environment(this)
+            % Test setting a new Python environment
+            % Note: Insert a valid path for the Python executable in your system
+            valid_python_path = 'C:\Users\TRA 3 PC0032\AppData\Local\Programs\Python\Python310\python.exe';  % Example path, change to a valid one for the test environment
```
teddychao, line 40: This can be simplified to `~/AppData/...`.
PsPM supports Win/Mac/linux. Previous build has considered Mac and Linux as much it can. Not sure if Mac and linux are considered in this occasion.

teddychao:
I am not sure I understand what test functions want to test, such as whether it wants to test if they are able to detect the python packages or what platforms they want to test.

teddychao (dismissed):
@MadniAbdulWahab Would you mind listing some goals of these test functions? Some specifications may be helpful, such as whether it wants to detect python path itself or it wants users to specify them. Thank you.

test/pspm_check_python_modules_test.m
MadniAbdulWahab: I don't think that will work because MATLAB does not directly recognize the tilde (~) as a shortcut for the home directory. We can try an indirect way like:
home_dir = getenv('USERPROFILE');  % This is for windows
if isempty(home_dir)
    % Fall back for macOS and Linux
    home_dir = getenv('HOME');
end

% Now construct the path with the determined home directory
src_path = fullfile(home_dir, 'GitHub', 'bachlab', 'PsPM', 'src');

test/pspm_check_python_test.m
MadniAbdulWahab: Actually the path is given by the user in the original function and the function works on it accordingly and sets or checks the environment. But when making test, I had to provided my path to test the function that's why I have given a windows path. You have to change the path accoring to your system.

teddychao:
Updated UI
Now users will have to specify the path of python before run heart data pre-processing with heartPy
image

image

dominikbach (approved):

PR #681, opened 2024-04-06
fix pspm_pp options call

### LINKED ISSUE #680: Bug in pspm_pp
### Summary
In line 66, pspm_options is called with the wrong argument value.

dominikbach:
Fixes #680  .

- fix a wrong argument value in the call of `pspm_options` from `pspm_pp`


teddychao (approved):

teddychao:
This is a minor correction so I merge it directly.

PR #684, opened 2024-04-10
Refactor `pspm_interpolate`

### LINKED ISSUE #673: Refactor `pspm_interpolate`
### Feature Description

1. Remove option to work on multiple data files at a time.
2. Remove option to work on several channels at a time: either work on all continuous channels (and write new file) or work on one channel only (and write to same file).
3. Use `pspm_load_channel` for single channel option, and `pspm_select_channels` for all channel option.
4. Change channel selector in GUI from `many` to `any`.

dominikbach:
Fixes #673 : refactors `pspm_interpolate`. 

1. Remove option to work on multiple data files at a time, as this is not used in other pre-processing functions.
2. Remove option to work on several channels at a time: either work on all continuous channels (and write new file) or work on one channel only (and write to same file).
3. Use `pspm_load_channel` for single channel option, and `pspm_select_channels` for all channel option.
4. Remove option to work on struct input - this is not supported by other PsPM functions and makes the code more clunky. (It is straightforward to deal with struct input as `pspm_load_data` gracefully handles this, but producing the output is more complex.)
5. Change channel selector in GUI is changed in reflection of step 2 from `many` to `any`. Also, file selector is changed to allowing just one file at a time.
6. Extrapolate handling is changed. Previously, if there were NaNs at beginning or end of file, `options.extrapolate` was forced to `1`.  However, then the option does not make sense, as it would always be enforced whenever it is relevant. Thus, this is no longer enforced, but a warning is still given that NaN remain in the data.
7. `pspm_options` is changed to deal with the larger number of methods allowed.
8. `pspm_load_channel` is changed to allow detecting whether a channel is a wave or events channel.


src/pspm_interpolate.m
```diff
@@ -1,51 +1,36 @@
...
+%   writes the result to the same file, or on all channels in the file and 
+%   writes a new file if the input is a file name. 
 % ● Format
-%   [sts, outdata] = pspm_interpolate(indata, options)
+%   [sts, outdata] = pspm_interpolate(numeric_array, options)
+%   [sts, outdata] = pspm_interpolate(filename, channel, options)
 % ● Arguments
-%          indata:  [struct/char/numeric] or [cell array of struct/char/numeric]
-%                   contains the data to be interpolated
+%          indata:  [char/numeric] contains the data to be interpolated
```
teddychao, line 12: Think the description of `indata` shall be updated to make it inline with above for avoiding confusion...

src/pspm_interpolate.m
```diff
@@ -65,238 +51,169 @@
...
-end
-if isempty(indata)
+elseif isempty(indata)
   warning('ID:missing_data', 'Input data is empty, nothing to do.');
   return;
 end
+
+if isnumeric(indata) 
+    if nargin >= 2
+        options = varargin{1};
```
teddychao, line 66: There is no issue for this situation, but I have a feeling that this refers to indata=numeric_array and `channel` does not exist in this case, so `nargin` can be only 1 or 2. ">2" is a situation that is unexpected.

src/pspm_interpolate.m
```diff
@@ -65,238 +51,169 @@
...
-    if inline_flag
-      dat = channel{k};
-    else
-      dat = channel{k}.data;
-    end
-    if numel(find(~isnan(dat))) < 2
+
+%% 2 work on data
+% 2.1 load data
+if method == 1
```
teddychao, line 109: I believe this `method` is different from `options.method` that is explained in help text. Maybe it is useful to explain the meaning of method 1, 2, 3 for future maintaining purposes.

src/pspm_interpolate.m
```diff
@@ -1,51 +1,36 @@
...
+%   writes a new file if the input is a file name. 
 % ● Format
-%   [sts, outdata] = pspm_interpolate(indata, options)
+%   [sts, outdata] = pspm_interpolate(numeric_array, options)
+%   [sts, outdata] = pspm_interpolate(filename, channel, options)
 % ● Arguments
-%          indata:  [struct/char/numeric] or [cell array of struct/char/numeric]
-%                   contains the data to be interpolated
+%          indata:  [char/numeric] contains the data to be interpolated
+%          channel: a single channel identifier accepted by pspm_load_channel
```
teddychao, line 13: I think accepting only one channel will be better because (1) it is simpler and can avoid bugs better (2) users can simply repeating this function in their script if they want to process multiple channels.

src/pspm_interpolate.m
```diff
@@ -65,238 +51,169 @@
...
+          if numel(xq) < 1
+            e_overlap = 0;
+            s_overlap = 0;
+          else
+            e_overlap = max(xq) > max(x);
+            s_overlap = min(xq) < min(x);
+          end
+          if s_overlap || e_overlap
+            if ~options.extrapolate
+              warning('ID:option_disabled', ...
```
teddychao, line 147: I am not sure I understand `option_disabled`, does this mean `options.extrapolate` must be true? It seems to be `option_invalid` which can be in line with `input_invalid`.

src/pspm_cfg/pspm_cfg_interpolate.m
teddychao, line 1: I suggest to add some help text to `cm`. For example, it can just let users know "only one channel will be processed under channel mode".

teddychao: And the channel action is renamed to `mode` here. I do not have a preference, but I think consistency can make users feel less confused about what they mean.
image

src/pspm_options.m
```diff
@@ -273,15 +273,15 @@
...
     options = autofill(options, 'extrapolate',            0,          1                 );
-    options = autofill(options, 'method',                 'linear',   {'pchip',...
+    options = autofill(options, 'method',                 'linear',   {'cubic', ...
+                                                                      'pchip',...
                                                                       'nearest', ...
                                                                       'spline', ...
                                                                       'previous', ...
                                                                       'next'}           );
 
     options = autofill(options, 'newfile',                0,          1                 );
```
teddychao, line 283: If this `newfile` is no longer used, I suggest to remove this line.

teddychao (approved):
Just made some minor suggestions. The updates to the functionality do not have issues to me. Thank you.

PR #707, opened 2024-05-08
Fix load channel

dominikbach:
Fixes an inconsistency in `pspm_load_channel`

- if the wrong channel type is loaded, a warning should be given, but processing should not be interrupted, i.e. the function should not return `sts < 1`


src/pspm_select_channels.m
```diff
@@ -40,7 +40,7 @@
     end
   case 'char'
     % in this case channel is specified as a char
-    if ~any(contains([{settings.channeltypes.type}, 'none', 'wave', 'events'], channel))
+    if ~any(startsWith([{settings.channeltypes.type}, 'none', 'wave', 'events'], channel))
```
teddychao, line 43: The update seems to allow some channels like `none_a`, which do not seem to be right? Would it be better to use `~any(startsWith([{settings.channeltypes.type}], channel)) || ~any(contains(['none', 'wave', 'events'], channel))`?

teddychao (approved):

teddychao:
Hi @dominikbach Thank you for this PR. Just a minor comment to the condition, which will not be very critical to the results as I understand.

src/pspm_select_channels.m
teddychao: | test channel type name | result |
|:--|:--|
| 'a' |  1 |
| 'none' | 0  |
| 'none_a' |  |
| "none" |  |
| 'scr' |  |
| 'scr_pp' |  |

teddychao (changes requested):
Please can we have a look at line 43 in pspm_select_channels? Thanks

dominikbach:
Sorted by naming all possible channel specifications explicitly.

teddychao (approved):

PR #710, opened 2024-05-21
fix options handling in pspm_extract_segments

dominikbach:
Fixes a bug in `pspm_extract_segments` by which a missing `options` input  led to a warning. Options are never mandatory, so this behaviour is changed now.


teddychao:
Tested and found the output as expected. Thank you.

PR #723, opened 2024-06-16
Fix issue 718 -- `display`

### LINKED ISSUE #718: Warning when running Display with full file path
### Summary

`pspm_display` works correctly with just the file name but throws a warning when called with full file path (from command line or GUI).

```
>> pspm_display('pspm_PP12.mat')
>> pspm_display('/Users/dominik/Documents/PublicData/Sandbox/pspm_PP12.mat')
Warning: The input to STR2FUNC "/Users/dominik/Documents/PublicData/Sandbox/pspm_PP12.mat" is not a valid
function name. This will generate an error in a future release. 
> In pspm_display (line 32) 
```

teddychao:
Fixes #718 

### Cause of the issue
The issue happened at line 32 in `pspm_display`. This is a line that is iteratively used when `pspm_display` is running. Initially, the `varargin` is set to be the input of `pspm_display`. In default, `varargin` is empty, so the line is not actually executed. In the next iteration, `varargin` becomes a multi-cell cell array, which is when `nargin` is no longer 0. This will not consider the situation that `pspm_display` actually has an input.
I am not fully clear what `gui_State.gui_Callback` is used for, since it is not really called much in this function. Therefore, I set a situation to identify whether it is processing the user's input as `varargin`, and this situation is determined in line 38 of the new version.

### Changes proposed in this pull request:
- This pull request updates the checking of input variable for display. If the `varargin{1}` refers to user's input, as a file name or a file path, it will no longer call `str2func`. The situation for determining the input is a file name or a file path is that the `varargin{1}` ends with `.mat`.

### DIFF
--- a/src/pspm_display.m
+++ b/src/pspm_display.m
@@ -29,7 +29,7 @@ gui_State = struct('gui_Name',       mfilename, ...
   'gui_LayoutFcn',  [], ...
   'gui_Callback',   []);
 if nargin && ischar(varargin{1})
-  gui_State.gui_Callback = str2func(varargin{1});
+  gui_State.gui_Callback = varargin{1};
 end
 
 if nargout
--- a/src/pspm_display.m
+++ b/src/pspm_display.m
@@ -2,12 +2,18 @@ function varargout = pspm_display(varargin)
 % ● Description
 %   pspm_display is the code for the GUI that is used to display different data from scr
 %   datafiles.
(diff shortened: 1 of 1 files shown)


teddychao:
It works now. The UI is ugly. I need to make a new pull request for the UI.
image

dominikbach (approved):

PR #729, opened 2024-06-20
add new HPR-reward model

### LINKED ISSUE #579: Implement new HPR model
### Feature Description

Implement new HPR model based on Xia, Liu et al. (in preparation).

This requires:
- writing a new function `pspm_bf_hpr_rew` (model & parameters to be asked from Ted)
- implement as a `modelspec` in `pspm_init`
- implement in GUI

dominikbach:
Fixes #579 .

- add new basis function `pspm_bf_hprf_rew`
- add new GLM to `pspm_init`
- add new GLM to GUI


src/pspm_cfg/pspm_cfg_glm_hp_rew.m
```diff
@@ -0,0 +1,50 @@
...
+% set callback function
+glm_hp_rew.prog = @pspm_cfg_run_glm_hp_rew;
+
+%% Basis function
+% HPRF
+hprf_rew        = cfg_const;
+hprf_rew.name   = 'HPRF_REW';
+hprf_rew.tag    = 'hprf_rew';
+hprf_rew.val    = {1};
+hprf_rew.help   = {'HPRF_REW without derivatives.'};
```
teddychao, line 34: I suggest to include the full name of "HPRF_REW" in the help text for people like me who do not have knowledge of the field but may have the intention to use it as a black box for own purposes.

src/pspm_cfg/pspm_cfg_glm_hp_rew.m
```diff
@@ -0,0 +1,50 @@
...
+vars = struct();
+vars.modality = 'HPR';
+vars.modspec = 'hp_fc';
+vars.glmref = { ...
+        ['Xia Y, Liu H, Kälin OK, Gerster S, Bach DR (under review). Measuring', ...
+         'human Pavlovian appetitive conditioning and memory retention.']};
+vars.glmhelp = '';
+
+% load default settings
+glm_hp_rew = pspm_cfg_glm(vars);
```
teddychao, line 19: There is a "Nuisance File" option, and it does not have to be specified, because on the current design. I am unsure if this is intentional or not. If it is an optional file, I would suggest to add "optional" in the help text. If it is mandatory, I suggest to report an error if it is not suggested. Currently it is not mandatory and can be left black for running the module.


image

src/pspm_cfg/pspm_cfg_glm_hp_rew.m
```diff
@@ -0,0 +1,50 @@
...
+
+% set correct name
+glm_hp_rew.name = 'GLM for HP (reward-conditioning)';
+glm_hp_rew.tag = 'glm_hp_rew';
+
+% set callback function
+glm_hp_rew.prog = @pspm_cfg_run_glm_hp_rew;
```
teddychao, line 26: There are errors that appear in `pspm_cfg_data_design_selector`, but they have been there before implementing #728 . I will mark the errors in that pull request #728.
image

teddychao (approved):

teddychao:
Minor comments regarding the text and UI. Critical errors will be put in #728 since they are more related to CFG functions than this new model.

src/pspm_cfg/pspm_cfg_glm_hp_rew.m
dominikbach: a generic stylistic issue for GLM, will be dealt with in future PR.

PR #734, opened 2024-06-24
Fix issue 733

### LINKED ISSUE #733: Spliting acqread and acqread_python from pspm_get_acq
### Summary
PsPM uses individual functions for specific features. Importing ACQ with python and importing ACQ with the classic method should be implemented with two different functions. The shared code should be put in the two functions.

teddychao:
Fixes #733.

- Split "get acq with python" and "get acq with classic methods".

Currently, "acq any version" (can be imported with `get_acq_python`) and "get acq equal to or earlier than version 3.9.0" (can be imported with `get_acq`) are split into two items in `pspm_init`.

Users can use `get_acq` to import acq files with the classic method or use `get_acq_python` to import acq files with the python package `bioread`.


teddychao:
## Screenshots

image

---

image

Note
"Bioread" is used as an item there because it wants to use the interface "pspm_cfg_selector_python". The title of this item has to be the name of the python package.

dominikbach (approved):

PR #743, opened 2024-07-16
Fix issue 730

### LINKED ISSUE #730: Improving `pspm_help`
### Feature Description

1. Change pspm_help such that the second output argument, arguments is a struct with field names corresponding to the argument names of the target function.
2. Add field "References" to the first argument. This should be a cell array of char, in which each char vector corresponds to oen reference. (`pspm_glm` is  a good test case).
3. Describe the rules by which function help is parsed. Currently, this throws errors for most functions but it is hard to fix these.
4. If possible, parse (or remove) matlab help references (as in `pspm_pupil_correct_eyelink`); if too complicated this can also be done manually on a case-by-case basis.

teddychao:
Fixes #730.

- Update `pspm_help` to make it able to read help text from source code.
- Update help text for functions.

After testing, almost all the functions can be read, with a few exceptions as following
- `pspm_get_timing`: Arguments are written in a different method.
- `pspm_jobman`: The symbol → is not read by `pspm_help`.
- `pspm_load_channel`: Arguments are not sorted conventionally.
- `pspm_load_data`: Arguments are not sorted conventionally.
- `pspm_process_illuminance`: There are second-level subfields.
- `pspm_pupil_pp_options`: Arguments are not sorted conventionally.
- `pspm_tam`: There are second-level subfields.


teddychao:
It can generate the struct now but have not sorted subfields. Will do next

image

teddychao:
Example

image

teddychao:
Example 2
image

dominikbach (approved):

PR #745, opened 2024-07-18
Multiple bug fixes

dominikbach:
Multiple bugs and improvements identified during beta testing.

- fixes bugs in channel index handling 
- fixes bugs in `pspm_init` (duplicated path removal, no warning if only required paths are already on the path)
- fixes a bug in a new functionality of `pspm_dcm`/`pspm_dcm_inv` (argument value `inf` is not recognised as `int` by `pspm_options`)
- fixes a bug in `pspm_extract_segments` (normalisation with missing values, `sr` for glm)
- fixes a bug in `pspm_glm` (wrong sample rate)
- improves `pspm_trim` (use `pspm_load_channel` and fix help text)
- improves `pspm_glm` (remove unnecessary and impossible orthogonalisation in edge case)
- catch an edge case in epoch definition (negative epoch onsets) in `pspm_get_timing` and `pspm_trim` 
- improve interpolation in `pspm_convert_hb2hp` and `pspm_resp_pp` by using nearest neighbour instead of linear extrapolation (including a new regression test)
- fixes a bug (missing index for a vector) in `pspm_multi2index`
- improve `pspm_pupil_pp` by using a different criterion for the decision on whether to combine pupils


src/pspm_init.m
teddychao, line 1: I fixed the issue of pspm_init in PR 746.

teddychao (approved):

PR #754, opened 2024-08-05
Fix issue 692 -- remove `varargout`

### LINKED ISSUE #692: Remove `varargout` in remaining functions
### Summary
All functions should return static output (i.e. no use of `varargout`). 
Some functions do not do this, including `pspm_overwrite` and `pspm_prepdata`. 
This requires checking the entire codebase where these functions are called.

## `pspm_overwrite`
1. This is an internal function with no user-specified arguments. Thus, the output argument `sts` is not needed, and we can settle on one static output argument.
2. There should be a third value for `overwrite` which signifies that the argument has not been set by the user.
3. In case of this third value, the GUI should be enabled in case the file exists.
4. `pspm_options` should use this third value as default. 
5. `pspm_options` should check the field `.overwrite` for all (and only for these) functions that create new files. 
6. Check all instances in PsPM where `pspm_overwrite` is called and ensure the output arguments are set appropriately.

## `pspm_prepdata`
1. This is an internal function but it checks a  user-supplied struct variable and so requires input checking and a `sts` output.
2. Thus, ensure a static output `[sts, data, newsr]`. 
3. Check all instances in PsPM where `pspm_prepdata` is called and ensure the output arguments are set appropriately.

## Other functions
Check all PsPM functions in the src folder and list all that use `varargout`.

teddychao:
Fixes #692.

- remove `varargout` in remaining functions, namely
  - `data_editor`
  - `display`
  - `ecg_editor`
  - `filtfilt`
  - `find_channel`
  - `jobman`
  - `load1`
  - `overwrite`
  - `prepdata`
  - `pulse_convert`
  - `pupil_pp`
  - `rev_dcm`
  - `rev_glm`
  - `tam`


dominikbach (approved):

teddychao:
Hi @dominikbach Thanks for approving this. Please kindly allow me a little time to update the description of these functions in documents.

PR #761, opened 2024-08-12
Fix issue 407

### LINKED ISSUE #407: A global variable refering to the GUI plotting window
One eventual issue that users might face is the interference between there plots and PsPM plots.

A way to resolve this would be to adjust the "plot-able" property of the PsPM's windows dynamically.
If every PsPM plotting functions had access to a global variable refering to the main figure to plot, they could switch it to "plot-able", do their things and then switch it back to "unplot-able".

teddychao:
Fixes #407.

The global option proposed from the issue description is not currently found useful.
All the four GUI functions can run without plotting on the GUI. Details are listed below.
Minor updates are provided to the GUI functions: making the tag name of the GUI to be the name of the corresponding function, i.e. the tag of the GUI "pspm_data_editor.fig" is now "pspm_data_editor", instead of "fgDataEditor". This is to make things consistent and avoid future confusion.

---

# Update 23 Sep 2024

## The current status

1. The "UIWait" can disable user's command line input (such as `plot(1:2)`) during running a function and expect users to do multiple GUI operations.  
  * In the develop branch, `pspm_display` and `pspm_review` will allow users to type in command line during running the GUI, but `pspm_ecg_editor` and `pspm_data_editor` do not allow. This means  `pspm_ecg_editor` and `pspm_data_editor` will never make the user's commands displaying in the GUI since they are disabled.

Screenshot 2024-09-23 at 10 05 19

## pspm_display
For `pspm_display`, after I call a new plot when `pspm_display` is running, the plot directly displays as a separate figure, rather than showing in the GUI. I think this is expected. Please see video below.

https://github.com/user-attachments/assets/51d7a428-5e94-4225-be61-b6adabd4641c

If I call the function by displaying a specific file, then plot something, it also works as expect

image

## pspm_review
For `pspm_review`, there are two situations
  * If the user has plotted something from the model, e.g. clicking "display all trials for one session", then the new plots will overwrite the plotted figures. Please check below

image
 
After running `plot([1,2,3])`
image

However, if the user has not plotted anything, and try to plot new information when they are only running the GUI of `pspm_review`, the new plot will run as a separate figure.
image


teddychao:
Preview

https://github.com/user-attachments/assets/e6ffdcf3-8180-4860-97c4-e6ada79c9de6


https://github.com/user-attachments/assets/499eb6ae-0b07-4309-a105-c7354f43b97a

dominikbach (approved):

PR #768, opened 2024-08-27
fix edge case in marker value import

### LINKED ISSUE #727: Check bdf import
### Discussed in https://github.com/bachlab/PsPM/discussions/719

<div type='discussions-op-text'>

<sup>Originally posted by **evl98** June  8, 2024</sup>
Hello,

I am trying to import and trim my SCR data from a biosemi device (.bdf file). However, whenever I try to trim the data it gives the following error:

Matrix index is out of range for deletion.
In file "C:\Users\0072494\OneDrive - Universiteit Utrecht\Documents\MATLAB\PsPM\PsPM_course_ESCAN2024\PsPM_v6.1.2\pspm_trim.m" (???), function "pspm_trim" at line 328.
In file "C:\Users\0072494\OneDrive - Universiteit Utrecht\Documents\MATLAB\PsPM\PsPM_course_ESCAN2024\PsPM_v6.1.2\pspm_cfg\pspm_cfg_run_trim.m" (???), function "pspm_cfg_run_trim" at line 45.

This happens regardless of the references I put in. The only way in which the data does trim, is when I only import my SCR channel, without the markers, and then use the time period from the start of the file as a reference. For this reason, I am suspecting there to be an issue with the import of the marker channel from my .bdf file; as even just using the start of the file as a reference gives this error whenever the marker channel is included.

Does anybody have an idea what is going wrong and how I could solve this?
Thanks!

Emily</div>

dominikbach:
Fixes #727, an edge case reported in https://github.com/bachlab/PsPM/discussions/719.

- generalise syntax for `[S(:).field]` for cases in which `numel(S(i).field) ~= 1` in a new function `pspm_struct2vec`
- call this in several import functions
- fix a bug that leads to `markerinfos.values` being char rather than num.

### DIFF
--- a/src/pspm_struct2vec.m
+++ b/src/pspm_struct2vec.m
@@ -1,0 +1,42 @@
+function v = pspm_struct2vec(S, field, warningtype)
+% ● Description
+%   pspm_struct2vec turns a numerical field in a multi-element structure
+%   array into a numerical vector. If in every element of the structure 
+%   array, the field has one element, this returns the same output as
+%   [S(:).field]. If fields ar empty or have more than one element, the
+%   output vector will be made to have the same number of elements as S,
+%   and a warning will be thrown. 
+% ● Format
+%   v = pspm_struct2vec(S, field, warningtype)
+% ● Arguments
+%   *           S : a structure array.
+%   *       field : name of a numerical field.
+%   * warningtype : ['marker' or 'generic'] Type of warning to the displayed
+%                   for the user.
+% ● Output
+%   *     v : numerical vector.
+% ● History
+%   Introduced in PsPM 7.0
+%   Written in 2024 Dominik R Bach (Uni Bonn)
+
+if nargin < 3 || strcmpi(warningtype, 'generic')
+    warningstr = 'Element';
+else
+    warningstr = 'Marker value';
+end
+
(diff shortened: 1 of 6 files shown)


src/pspm_struct2vec.m
```diff
@@ -0,0 +1,42 @@
...
+%   Written in 2024 Dominik R Bach (Uni Bonn)
+
+if nargin < 3 || strcmpi(warningtype, 'generic')
+    warningstr = 'Element';
+else
+    warningstr = 'Marker value';
+end
+
+% for loops are faster than arrayfun
+n = numel(S);
```
teddychao, line 29: I tested this function with two-dimensional structure array, and I can see it works well with this function. The only thing is, this function will provide a 1xN vector `v`, where `N` denotes the number of elements in the structure array. If it is a two dimensional structure array with size `p x q`, the output will be `1 x p*q` vector.

teddychao (approved):

src/pspm_struct2vec.m
teddychao: Testing code:

``
field = 'f';
value = {'some text','a';
[10, 20, 30],'b';
magic(5),3};
S = struct(field,value);
v = pspm_struct2vec(S, 'f')

PR #771, opened 2024-08-29
Fix `pspm_extract_segments`

dominikbach:
Fixes a bug detected in beta testing.

- save events in `pspm_glm`
- correct computation of events in case of edge cases in `pspm_glm`
- add time stamp to the `glm` struct for troubleshooting
- correct `pspm_extract_segments` when time units are markers and input is a GLM


src/pspm_glm.m
```diff
@@ -491,6 +494,7 @@
 %% 12 collect data & regressors for output model
 glm.input.data    = y;
 glm.input.sr      = sr;
+glm.input.events  = events;
```
teddychao, line 497: I can add the description of `glm` to `Outputs` in help text if useful

teddychao:
In `pspm_extract_segments_test`, at line 208, `pspm_extract_segments` expects that `glm` has the field `glm.input.events`, as this is used by line 158 in `pspm_extract_segments`. In the current version of `pspm_extract_segments_test`, there is no `glm.input.events`. I do not know if the test data shall be updated?

image

teddychao (approved):

PR #786, opened 2024-09-30
758 pspm_expand_epochs

### LINKED ISSUE #758: Create a function to expand epochs

4gwe:
Fixes #758  .

-  made pspm_expand_epochs
-  added to to pspm_remove
-


src/pspm_expand_epochs.m
```diff
@@ -0,0 +1,147 @@
...
+    % missing epochs
+    case 'missing_ep'
+        % Directly expand the given epochs
+        [sts, ep_exp] = expand(epoches, expansion);
+        return;
+
+    % missing epoches file
+    case 'missing_ep_file'
+        % Load missing epochs from file
+        data = load(epoches);  % Load the file without specifying a variable
```
dominikbach, line 33: Epochs should always be loaded through pspm_get_timing - this also checks the content of the file.

src/pspm_expand_epochs.m
```diff
@@ -0,0 +1,147 @@
+function [sts, ep_exp] = pspm_expand_epochs(epoches, expansion, options)
```
dominikbach, line 1: I would suggest using the `varargin` syntax.

src/pspm_expand_epochs.m
```diff
@@ -0,0 +1,147 @@
...
+
+function [sts, ep_exp] = expand(ep, expansion)
+% Helper function to expand epochs by the specified pre and post times
+% and merge overlapping epochs.
+% Also ensures that no epoch starts before time 0.
+
+% Initialize status
+sts = -1;
+
+% Check if epochs matrix and expansion vector are valid
```
dominikbach, line 118: Since this is needed for all modes of the function, I would check it in the beginning of the main function, before anything is loaded from file (faster and easier to read).

src/pspm_expand_epochs.m
```diff
@@ -0,0 +1,147 @@
...
+    return;
+end
+
+% Expand epochs
+pre = expansion(1);
+post = expansion(2);
+expanded_epochs_temp = [ep(:,1) - pre, ep(:,2) + post];
+
+% 
+% Ensure that the start of any epoch is not negative
```
dominikbach, line 130: 0 is correct.

src/pspm_expand_epochs.m
```diff
@@ -0,0 +1,147 @@
...
+expanded_epochs_temp(expanded_epochs_temp(:,1) < 0, 1) = 0; % or <1???
```
dominikbach, line 131: Probably we should put this code into pspm_get_epochs (if the functionality does not exist there already), because it is needed in several functions. 

then just call [sts, epochs] = pspm_get_timing('epochs', epochs)

src/pspm_remove_epochs.m
```diff
@@ -12,14 +12,17 @@
 %                 be in seconds. This parameter is passed to pspm_get_timing().
 %   * timeunits:  timeunits of the epochfile.
 %   ┌───options
-%   └─.channel_action:
+%   └─.channel_action: [pre, post]
```
dominikbach, line 15: why is the channel action changed?

src/pspm_expand_epochs.m
```diff
@@ -0,0 +1,185 @@
...
+            [ests, ep_exp] = expand(epochs.epochs, expansion);
+            if ests == -1
+                error('Failed to expand epochs.');
+                return;
+            end
+
+            % Save expanded missing epoch to a new file with 'e' prefix
+            [pathstr, name, ext] = fileparts(filename);
+            output_file = fullfile(pathstr, ['e' name ext]);
+            save(output_file, 'ep_exp'); % should i save it as epoch???
```
dominikbach, line 69: Yes, the variable in the output file shold be named "epochs", too.

dominikbach (changes requested):
some functions return sts == -1 if unsuccessful, others sts == 0. So to be most general, it is best practice to check for unsuccessful downstream functions with `if sts < 1`

dominikbach (approved):

PR #793, opened 2024-10-23
Update import_eyelink.m

ZhangSihui999:
correct import for both eyes

Fixes # .

- 
- 
-


4gwe:
hi @ZhangSihui999,

Thank you so much for your pull request! We appreciate the effort you put into this.

I have a quick question: does this fix address a general issue, or is it specific to the data you’re working on? If possible, could you share some sample data with us? This would be super helpful for us to test the fix on our end.

PR #796, opened 2024-11-04
787 pspm_downsample

### LINKED ISSUE #787: Problem with downsampling missing values in `pspm_dcm`.
In what is currently line 404 of `pspm_dcm`, the function `downsample` is called with a potentially non-integer sample rate, which will throw an error. The function `resample` (used in `pspm_prepdata`) would work here, but only if the signal processing toolbox is installed.

Suggested workaround:
- `pspm_downsample` becomes the main function for handling downsampling, and integrates lines 154-178 of `pspm_prepdata`
- `pspm_prepdata` and `pspm_dcm` call `pspm_downsample` for downsampling. 
- create tests for `pspm_downsample` 

_Originally posted by @dominikbach in https://github.com/bachlab/PsPM/discussions/719#discussioncomment-10870203_

4gwe:
Fixes  #787 .

- pspm_downsample got the resample feature from (pspm_prepdata)
- pspm_prepdata uses  pspm_downsample
-  pspm_downsample uses pspm_dcm (implementation in progress).
- created pspm_downsample_test

### DIFF
--- a/src/pspm_downsample.m
+++ b/src/pspm_downsample.m
@@ -35,24 +35,27 @@ end
 
 freqratio = sr/sr_down;
 
-if settings.signal
-    % this allows real sr ratios but requires integer initial and final sr
+if freqratio == ceil(freqratio) % NB isinteger might not work for some values
+    % this gives the same output as the signal processing toolbox function
+    % downsample(data, freqratio)
+    data = data(1:freqratio:end); % from old pspm_downsample
+    newsr = sr_down;
+    sts = 1;
+elseif settings.signal
+    % this uses a filter and is therefore not used for integer sampling
+    % ratio. It allows real sr ratios but requires integer initial and final sr
     if sr == floor(sr) && sr_down == floor(sr_down)
         data = resample(data, sr_down, sr);
         newsr = sr_down;
     else
-        % use a crude but very general way of getting to integer numbers
+        % use a crude but very general way of getting to integer numbers by
+        % changing the new sampling rate
         altsr = round(sr);
         altdownsr = round(sr_down);
         data = resample(data, altdownsr, altsr);
         newsr = sr * altdownsr/altsr;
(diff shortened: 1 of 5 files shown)


src/pspm_downsample.m
```diff
@@ -23,10 +24,42 @@
...
+%   warning('Frequency ratio must be integer.'); return
 end
 %% 3 Performing downsampling
-data = data(freqratio:freqratio:end);
-sts = 1;
+
+freqratio = sr/sr_down;
+
+
+if freqratio == ceil(freqratio) % NB isinteger might not work for some values
```
dominikbach, line 35: I'd suggest exchanging the first and second case, i.e. start with `if settings.signal`. This way, we can be sure that an in-built function handles most of the cases, even when frequency ratio is integer.

src/pspm_downsample.m
```diff
@@ -23,10 +24,42 @@
 %% 2 Check input arguments
 if nargin < 2
   warning('Not enough input arguments.'); return
-elseif floor(freqratio) ~= freqratio
-  warning('Frequency ratio must be integer.'); return
+% elseif floor(freqratio) ~= freqratio
```
dominikbach, line 27: Please delete the commented out lines.

src/pspm_downsample.m
```diff
@@ -1,9 +1,9 @@
-function [sts, data] = pspm_downsample(data, freqratio)
+function [sts, data, newsr] = pspm_downsample(data, sr,sr_down)
 % ● Description
```
dominikbach, line 2: Please change help text. This function is now used more generally, not just for users without signal processing toolbox.

dominikbach (changes requested):

src/pspm_downsample.m
4gwe: But will the resampled data not be a bit off (in the case of integer ratios)?

dominikbach (approved):

PR #808, opened 2024-12-15
Changes to pspm_scr_pp

4gwe:
Fixes # .

-  added pspm_expand_epochs to pspm_scr_pp
-  added pspm_logical2epochs and pspm_epochs2logical to pspm_scr_pp
-  deleted internal function filter_to_epochs
- added a control for negative epochs intervals in pspm_expand_epochs


src/pspm_expand_epochs.m
```diff
@@ -117,6 +117,11 @@
 pre = expansion(1);
 post = expansion(2);
 expanded_epochs_temp = [epochs(:,1) - pre, epochs(:,2) + post];
+
+% delets epochs whos difference is negative or 0
```
dominikbach, line 121: epoch diff cannot become negative by subtracting a positive number from the start and a adding a positive number  to the end. This could only happen if pre/post are negative (which happens because of line 179 in pspm_scr_pp?)

I'd suggest you add an input check on the `expansion`  input argument; the lines 122-123 here can probably be removed.

dominikbach (changes requested):

dominikbach (approved):

PR #814, opened 2025-01-07
fix smrx and associated bugs

### LINKED ISSUE #811: Import smrx error
### Summary
The user attempts to import a smrx file. After typing the import settings below, there is an error saying there is no subfield `fileinfo.chaninfo.kind`.

### Steps to Reproduce
image

### Technical Info
* PsPM version: develop
* MATLAB version: 2024a

dominikbach:
Fixes #811 .

- fix a bug in smrx import when searching for channel
- fix a bug in smrx import that would turn all warnings off and not on again
- fix a bug in import that would prevent a file from being overwritten when the user is asked via dialogue


teddychao (approved):
Looks good to me. Thank you.

PR #836, opened 2025-02-25
822 pspm pupil pp unit check

### LINKED ISSUE #822: Add unit check for pupil data in `pspm_pupil_pp`
The function `pspm_pupil_pp` cannot process pupil data in ratio units. A check is needed to detect ratio units and handle them appropriately.

4gwe:
Fixes #822 .

- Added a check for allowed pupil data units.

### DIFF
--- a/src/pspm_pupil_pp.m
+++ b/src/pspm_pupil_pp.m
@@ -153,6 +153,16 @@ if sts_load < 1, return, end
 [sts_load, data,infos, pos_of_channel(1)] = pspm_load_channel(alldata, options.channel, 'pupil');
 if sts_load ~= 1, return, end
 
+% Check for invalid pupil data units 
+allowed_units = {'mm', 'cm', 'dm', 'm', 'km', 'in', 'inches'};
+if ~ismember(data.header.units, allowed_units)
+    warning('pspm_pupil_pp:InvalidUnit', ...
+          'Unsupported pupil data unit: %s. Allowed units are: %s', ...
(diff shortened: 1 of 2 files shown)


src/pspm_pupil_pp.m
```diff
@@ -225,6 +245,8 @@
 [sts, out_id] = pspm_write_channel(fn, smooth_signal, options.channel_action, o);
 outchannel = out_id.channel;
 
+if con_flag == 1 , warning('At least one channel was converted to "mm". See details above.'); end
```
dominikbach, line 248: What does "see details above" refer to?

src/pspm_pupil_pp_options.m
```diff
@@ -16,9 +16,9 @@
 %   │ // Allowable values criteria
 %   │ 
 %   ├─PupilDiameter_Min:  Minimum allowable pupil size. Pupil values less than
```
dominikbach: Suggest: "Minimum allowable pupil size in mm"
and same in line 20

dominikbach (changes requested):

src/pspm_pupil_pp.m
4gwe: If there is a conversion, a warning will be displayed (one per conversion). However, all the information about pupil preprocessing will still be shown afterward, which can be a lot. This serves as a reminder.
**Like this:**
image

dominikbach (approved):

PR #853, opened 2025-09-04
epochs edgecase fix in pspm_logical2epochs and more test

4gwe:
Fixes an edge case in conversion from logical index to epochs.

- the epoch indices should not have a larger index than the number of data points.
- 
-


dominikbach (approved):

dominikbach (changes requested):
Just realised that if `sr ~=1`, then this is not an issue. So I would suggest adding this as an `else` statement the last if-clause?

dominikbach (approved):

PR #866, opened 2025-11-05
replace nan-functions

gjheij:
Fixed `nanmean` in `pspm_extract_segments.m`. `nanmean` was removed from core MATLAB a while back. In newer releases (including R2025a), the recommended replacement is to use the mean function with the `'omitnan'` flag.

before:
```matlab
  segments{c}.mean = nanmean(m, 1);
  segments{c}.std = nanstd(m, [], 1);
```

now:
```matlab
  segments{c}.mean = mean(m, 1, 'omitnan');
  segments{c}.std = std(m, [], 1, 'omitnan');
```


dominikbach (approved):

dominikbach (approved):

dominikbach:
merged via #870

PR #900, opened 2026-05-13
Non-linear model defaults

dominikbach:
Fixes # .

Changed non linear model default in line with de Vries et al 2026) preprint. 

- change default filter to unidirectional 
- change default flexible response dispersion to constrained, as simplest option for downstream analysis
- integrate constrained and the hidden upper_constrained option

Please test with tutorial data.


gjheij (commented):
Looks good to me

dominikbach:
aSCR offset was interchangeably taken from two different sources (`pspm_init` and `pspm_options`) which had different default values. Now removed from `pspm_init` and only set in DCM options. Help text adjusted as well.

dominikbach:
Current status:

f_SCR requires sigma_offset, which is currently passed via settings in line 54, and this setting field was removed. We need to create a different way of passing it. The easiest solution might be a field settings.temp  which is set in pspm_inv  (and removed at the end of the script) and read in f_SCR.

gjheij (approved):

PR #910, opened 2026-08-13
Support segmented AcqKnowledge imports

### LINKED ISSUE #848: Support for importing exported segments from Biopac AcqKnowledge (*.mat)
### Feature Description

Currently, the function `pspm_get_acqmat` does not support .mat files **exported as segments** from Biopac AcqKnowledge. 

### Feature Request:
Update `pspm_get_acqmat` to support exported segmented .mat files.

### Relevant Code

`pspm_get_acqmat`  around line 63

4gwe:
Fixes #848 

## Changes proposed in this pull request:

Support segmented AcqKnowledge MAT imports with non-zero start_sample.

The import no longer aborts for non-zero start offsets. Instead, PsPM warns about the offset and stores start_sample together with its unit in sourceinfo.

### DIFF
--- a/src/pspm_get_acqmat.m
+++ b/src/pspm_get_acqmat.m
@@ -29,12 +29,21 @@ if isempty(settings)
   pspm_init;
 end
 sts = -1;
-sourceinfo = [];
+sourceinfo = struct();
 
 % load data
 % -------------------------------------------------------------------------
 inputdata = load(datafile);
 
+% store original recording offset
+sourceinfo.start_sample = inputdata.start_sample;
+sourceinfo.start_sample_unit = inputdata.isi_units;
+
+if inputdata.start_sample ~= 0
+  warning('ID:acqmat_nonzero_start', ...
+    'AcqKnowledge data starts at %g %s relative to the original recording.', ...
+    inputdata.start_sample, strtrim(inputdata.isi_units));
+end
 
 % extract individual channels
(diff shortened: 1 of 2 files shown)


dominikbach:
Quick clarification questions. 
1. Does Acknowledge allow digital channels recorded as time stamps? And if yes, are these time stamps recorded wrt to the original recording start? Should they then be corrected wrt to wave recording start?
2. Is the offset the same for all wave channels, or could they have different offset?

dominikbach (changes requested):
see my questions.

4gwe:
> Quick clarification questions.
> 
> 1. Does Acknowledge allow digital channels recorded as time stamps? And if yes, are these time stamps recorded wrt to the original recording start? Should they then be corrected wrt to wave recording start?
> 2. Is the offset the same for all wave channels, or could they have different offset?

1. When exporting in the MATLAB format, all channels need to be represented as waveform data, including events, so events have to be converted to a waveform channel before export. AcqKnowledge 5/6 can also export events separately using "Export for MATLAB/Octave with Events", which creates a separate `Markers.m` file that can be used to import the markers into the MATLAB workspace. I could not find sufficient information to include support for this in `pspm_get_acqmat`.

2. For wave channels, the *.mat contains a single 'start_sample' for the entire data array, so the offset for all exported channels.

dominikbach (approved):
Sounds good to me.

